PR #16, opened 2015-02-09
Small improvements

lynxis:
Hi,
here are some improvements.
They are tested somehow, but they need additional testing. I would like have some automatic tests :).
I'll use this branch for my wifi at home for a week now, but this wouldn't cover all cases.

Best,
lynxis

### DIFF
--- a/broker/l2tp_broker.py
+++ b/broker/l2tp_broker.py
@@ -390,6 +390,7 @@ class Tunnel(gevent.Greenlet):
     self.limits = Limits(self)
     self.sessions = {}
     self.next_session_id = 1
+    self.id = -1
     self.keep_alive()
 
   def setup(self):
--- a/broker/l2tp_broker.py
+++ b/broker/l2tp_broker.py
@@ -772,7 +772,7 @@ class TunnelManager(object):
       return
 
     # Execute the registered hook
-    logger.debug("Executing hook '%s' via script '%s'..." % (name, script))
+    logger.debug("Executing hook '%s' via script '%s %s'." % (name, script, str([str(x) for x in args])))
     try:
       gevent_subprocess.call([script] + [str(x) for x in args])
     except:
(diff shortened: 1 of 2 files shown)


mitar:
Feel free to suggest how to do tests. :-) Travis CI sadly does not allow LXC containers inside, that one could create multiple virtual boxes and networking stack to test things.

broker/l2tp_broker.py
```diff
@@ -59,7 +59,7 @@
   # Message data (with length prefix)
   cs.PascalString("data"),
   # Pad the message so it is at least 12 bytes long
-  cs.Padding(lambda ctx: max(0, 6 - len(ctx["data"]))),
+  cs.Optional(cs.Padding(lambda ctx: max(0, 6 - len(ctx["data"])))),
```
kostko: Can you also remove the code around line 525 (method `Tunnel._run`) that adds padding before parsing if there is none, as now this is probably not necessary and your fix is actually much nicer.

broker/l2tp_broker.py
```diff
@@ -772,7 +773,7 @@ def hook(self, name, *args):
       return
 
     # Execute the registered hook
-    logger.debug("Executing hook '%s' via script '%s'..." % (name, script))
+    logger.debug("Executing hook '%s' via script '%s'... %s" % (name, script, str([str(x) for x in args])))
```
kostko: Can you move the last `%s` to be placed next to the script name? So something like `via script '%s %s'...`.

kostko:
Looks good, just address the above comments and I will merge it. Thanks!

mitar:
@lynxis, please address the pull request feedback so that we can merge this in. ;-)

lynxis:
done.

kostko:
Ok, looks good now, thanks, I will merge it.

> Travis CI sadly does not allow LXC containers inside, that one could create multiple virtual boxes and networking stack to test things.

But CircleCI allows running your own Docker containers ;-) So perhaps some tests could be done this way?

mitar:
> But CircleCI allows running your own Docker containers ;-) So perhaps some tests could be done this way?

Are they free for open source projects?

kostko:
Yes, they even give you more containers so you can run more parallel builds if you are an open source project.

PR #17, opened 2015-03-20
Adding logging of script outputs.

max-b:
Hey so this ended up just being a small useful tool for us to be able to get logging output of any of the hook scripts.

### DIFF
--- a/broker/l2tp_broker.py
+++ b/broker/l2tp_broker.py
@@ -796,10 +796,10 @@ class TunnelManager(object):
       stdout, stderr = script_process.communicate()
 
       if stdout:
-          logger.debug("script stdout:")
+          logger.debug("Hook '%s' script stdout:" % script)
           logger.debug(stdout)
       if stderr:
-          logger.warning("script stderr:")
+          logger.debug("Hook '%s' script stderr:" % script)
           logger.warning(stderr)
     except:
       logger.warning("Failed to execute hook '%s'!" % script)
--- a/broker/l2tp_broker.py
+++ b/broker/l2tp_broker.py
@@ -92,9 +92,6 @@ LimitMessage = cs.Struct("limit",
   cs.PascalString("data")
 )
 
-# For logging scripts
-script_logger = open(config.get('broker', 'script_log'), 'a+')
-
 LIMIT_TYPE_BANDWIDTH_DOWN = 0x01
 
 # Overhead of IP and UDP headers for measuring PMTU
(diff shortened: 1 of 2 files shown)


mitar:
Shouldn't we just make it so that all logging goes to `tunneldigger-broker.log`? So that we capture the script output and put it into `tunneldigger-broker.log` as well?

max-b:
Yeah I generally agree with you on this. This was something we originally wrote quite a few months ago, but if I remember correctly there was an issue redirecting stderr and stdout from the scripts to the logging module.

Doing some searching it looks like either of these might work:
http://stackoverflow.com/questions/9410760/redirect-stdout-to-logger-in-python
or
http://stackoverflow.com/questions/19425736/how-to-redirect-stdout-and-stderr-to-logger-in-py

I'll take a look when I get a chance.

max-b:
Ok this looks like it's working on our end. Lemme know if there's anything else?

mitar:
Ping?

broker/l2tp_broker.py
```diff
@@ -768,7 +768,15 @@ def hook(self, name, *args):
     # Execute the registered hook
     logger.debug("Executing hook '%s' via script '%s %s'." % (name, script, str([str(x) for x in args])))
     try:
-      gevent_subprocess.call([script] + [str(x) for x in args])
+      script_process = gevent.subprocess.Popen([script] + [str(x) for x in args], stdout=gevent.subprocess.PIPE, stderr=gevent.subprocess.PIPE)
+      stdout, stderr = script_process.communicate()
+
+      if stdout:
+          logger.debug("script stdout:")
```
kostko: Can you instead use something like `"Hook '%s' script stdout:"`? Since `communicate` may yield, the output could appear after some other hooks have been executed and then you can't know which script produced the output.

Similar for stderr.

max-b:
Yeah that makes sense. I haven't had a chance to test this on an actual tunnel broker, but it seemed like I would've really had to make a dumb mistake to miss something here :)  That being said, when I get a chance I can make sure that it works as expected.


Thanks!

max-b:
@kostko - any chance you'd want to take another look at this now?

broker/l2tp_broker.py
```diff
@@ -792,7 +792,15 @@ def hook(self, name, *args):
     # Execute the registered hook
     logger.debug("Executing hook '%s' via script '%s %s'." % (name, script, str([str(x) for x in args])))
     try:
-      gevent_subprocess.call([script] + [str(x) for x in args])
+      script_process = gevent.subprocess.Popen([script] + [str(x) for x in args], stdout=gevent.subprocess.PIPE, stderr=gevent.subprocess.PIPE)
```
kostko: This will not work with legacy `gevent_subprocess`. Can you replace `gevent.subprocess` with `gevent_subprocess` (we detect the proper module at the top)?

kostko: Also please break the line so each argument is in its own line.

broker/l2tp_broker.py
```diff
@@ -792,7 +792,15 @@ def hook(self, name, *args):
...
+      stdout, stderr = script_process.communicate()
+
+      if stdout:
+          logger.debug("Hook '%s' script stdout:" % script)
```
kostko: This code file is still using 2 spaces to indent, so you should follow this style. I agree that we should migrate to 4 spaces in the future and make the whole thing PEP8 compliant. But at the moment, please make it the same style as other code.

kostko:
Looked at it again and added some minor comments. Please address them and then it should be good.

max-b:
Thanks for the feedback @kostko. Look okay now?

kostko:
LGTM, merging.

max-b:
Awesome - thanks for the hand holding!

PR #19, opened 2015-05-29
Added -I option for binding to a specific interface.

max-b:
I think someone already made a request re: this commit, but I can't seem to find it. I can edit this pull request if someone would remind me?

### DIFF
--- a/client/l2tp_client.c
+++ b/client/l2tp_client.c
@@ -189,6 +189,9 @@ void context_free(l2tp_context *ctx);
 static l2tp_context *main_context = NULL;
 static asyncns_t *asyncns_context = NULL;
 
+/* Force the tunnel to go over a certain interface */
+char *force_iface = 0;
+
 time_t timer_now()
 {
   struct timespec ts;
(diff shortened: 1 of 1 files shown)


client/l2tp_client.c
```diff
@@ -189,6 +189,9 @@ void context_free(l2tp_context *ctx);
 static l2tp_context *main_context = NULL;
 static asyncns_t *asyncns_context = NULL;
 
+/* Force the tunnel to go over a certain interface */
+char *force_iface = 0;
```
kostko: It would be better to store this in the `l2tp_context`, as the use of global state should really be minimized. This enables one to have multiple context bound to different interfaces within the same application (not needed at the moment, but it makes for a nicer API).

kostko:
Just move the `force_iface` state to `l2tp_context` and set it up in `context_new`. Otherwise LGTM.

max-b:
Ok I just compiled this and it looks like it's working as expected. Thanks for the feedback.

client/l2tp_client.c
```diff
@@ -282,6 +285,8 @@ l2tp_context *context_new(char *uuid, const char *local_ip, const char *broker_h
   ctx->tunnel_id = tunnel_id;
   ctx->hook = hook ? strdup(hook) : NULL;
 
+  ctx->force_iface = force_iface;
```
kostko: Can you make a copy of the `force_iface` string - similarly to how it is done for `hook`? And then don't forget to also free `ctx->force_iface` in `context_free`.

max-b:
Cool thanks for walking me through this. Still a little early over here and hadn't had any coffee yet...

kostko:
No problem, thanks for your pull request! ;-) Looks good now.

PR #20, opened 2015-09-26
Fixing string length argument in setsockopt call as per @juul's catch.

max-b:
@Juul mentioned that setsockopt expects a null terminated string, and that `strlen(ctx->force_iface)` is only going to give the length of the interface minus the null-terminator, so we want to make sure that we include that last null-termination byte.


kostko:
Thanks!

PR #21, opened 2015-10-06
Updated requirements and configurable kernel modules check

lynxis:


kostko:
Did you verify that the broker works with the updated requirements?

lynxis:
yes. it works.

PR #23, opened 2015-10-15
Add sequence numbers to keep-alive packets

lynxis:
Some ISP killing duplicated packets. A sequence number
prevents this

### DIFF
--- a/broker/l2tp_broker.py
+++ b/broker/l2tp_broker.py
@@ -393,6 +393,14 @@ class Tunnel(gevent.Greenlet):
     self.next_session_id = 1
     self.id = -1
     self.keep_alive()
+    self.__keepalive_seqno = 0
+
+  @property
+  def _keepalive_seqno(self):
+    self.__keepalive_seqno += 1
+    if self.__keepalive_seqno >= (2**32):
+        self.__keepalive_seqno = 0
+    return self.__keepalive_seqno
 
   def setup(self):
     """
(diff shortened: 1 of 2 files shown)


kostko:
There are ISPs that do this? Why?

Anyway, the patch looks good, just please add some comments explaining why a sequence number is needed in a keepalive message.

lynxis:
> There are ISPs that do this? Why?
I would like to know that. cable ISP and mobile network operators doing such miserable things.

kostko:
I would add that to both, the broker and the client in the place where the keepalive message is generated.

lynxis:
done

broker/l2tp_broker.py
```diff
@@ -393,6 +393,14 @@ def __init__(self, manager, port):
     self.next_session_id = 1
     self.id = -1
     self.keep_alive()
+    self.__keepalive_seqno = 0
```
mitar: Waw. Why `__`? This is really unnecessray.

broker/l2tp_broker.py
```diff
@@ -393,6 +393,14 @@ def __init__(self, manager, port):
...
+
+  @property
+  def _keepalive_seqno(self):
```
mitar: Also it is great to have verbose names. This could be `nextSequenceNumber` and not reason for a property.

mitar: (So that it is clear that there are side-effects.)

mitar:
I updated it in `master`.

PR #25, opened 2015-10-15
Add usage selection to broker discovery

lynxis:
This is an RFC.

It's based on commits from Christian Mehlis.
* implement broker selection in a generic way
* Client ask for an usage and select the broker with the smallest value.

It's a way to implement load balancing over multiple hosts. The broker use the max tunnel configuration to calculate it's usage. The usage itself is a 1 Byte metric sent to the client.
The client decide then to whom it will connect.

Before I go refactoring this, my question is:
Do we want this feature (not this exact code) to be upstream?
Is the usage a optional feature? Configurable? Or do we want all new servers supporting this, but still be compatible to old clients?

* **This needs testing**
* **This needs refactoring into smaller commits**
* **This pullrequest break the client**

### DIFF
--- a/client/l2tp_client.c
+++ b/client/l2tp_client.c
@@ -92,6 +92,7 @@ enum l2tp_ctrl_type {
   CONTROL_TYPE_PMTUD_ACK = 0x07,
   CONTROL_TYPE_REL_ACK   = 0x08,
   CONTROL_TYPE_PMTU_NTFY = 0x09,
+  CONTROL_TYPE_USAGE     = 0x0A,
 
   // Reliable messages (0x80 - 0xFF)
   CONTROL_TYPE_LIMIT     = 0x80,
@@ -114,6 +115,7 @@ enum l2tp_limit_type {
 };
 
 enum l2tp_ctrl_state {
+  STATE_GET_USAGE,
   STATE_GET_COOKIE,
   STATE_GET_TUNNEL,
   STATE_KEEPALIVE,
@@ -153,6 +155,8 @@ typedef struct {
   int fd;
   // Tunnel state
   int state;
+  // Usage
+  uint16_t usage;
   // Cookie
   char cookie[8];
   // Netlink socket
@@ -191,6 +195,7 @@ typedef struct {
 
   // Last keepalive and timers
   time_t last_alive;
+  time_t timer_usage;
   time_t timer_cookie;
   time_t timer_tunnel;
   time_t timer_keepalive;
@@ -379,9 +384,11 @@ int context_reinitialize(l2tp_context *ctx)
   if (ctx->broker_resq)
     asyncns_cancel(asyncns_context, ctx->broker_resq);
   ctx->broker_resq = NULL;
+  ctx->usage = -1;
 
   // Reset relevant timers
   time_t now = timer_now();
+  ctx->timer_usage = now;
   ctx->timer_cookie = now;
   ctx->timer_tunnel = now;
   ctx->timer_keepalive = now;
@@ -492,6 +499,19 @@ void context_process_control_packet(l2tp_context *ctx)
 
   // Check packet type
   switch (type) {
+    case CONTROL_TYPE_USAGE: {
+      if (ctx->state == STATE_GET_USAGE) {
+        // broker usage information
+        ctx->usage = parse_u16(&buf);
+        syslog(LOG_DEBUG, "Broker usage: %p %s %u\n", ctx, ctx->broker_hostname, ctx->usage);
+
+        // Mark the connection as being available for later establishment
+        ctx->standby_available = 1;
+        ctx->timer_cookie = timer_now();
+        ctx->state = STATE_GET_COOKIE;
+      }
+      break;
+    }
     case CONTROL_TYPE_COOKIE: {
       if (ctx->state == STATE_GET_COOKIE) {
         if (payload_length != 8)
@@ -947,7 +967,8 @@ void context_process(l2tp_context *ctx)
             syslog(LOG_ERR, "Failed to connect to remote endpoint - check WAN connectivity!");
             ctx->state = STATE_REINIT;
           } else {
-            ctx->state = STATE_GET_COOKIE;
+            ctx->timer_usage = timer_now();
+            ctx->state = STATE_GET_USAGE;
           }
           asyncns_freeaddrinfo(result);
           ctx->broker_resq = NULL;
@@ -961,6 +982,15 @@ void context_process(l2tp_context *ctx)
         ctx->state = STATE_REINIT;
         return;
       }
+      break;
+    }
+    case STATE_GET_USAGE: {
+      // Get usage information if available
+      if (!is_timeout(&ctx->timer_usage, 2))
+        context_send_packet(ctx, CONTROL_TYPE_USAGE, "UUUUUUUU", 8);
+      else
+        // Time out. The broker might not support usage. Ignore it.
+        ctx->state = STATE_GET_COOKIE;
       break;
     }
     case STATE_GET_COOKIE: {
@@ -1101,6 +1131,23 @@ typedef struct {
 } broker_cfg;
 #define MAX_BROKERS 10
 
+int broker_selector_usage(broker_cfg *brokers, int broker_cnt, int ready_cnt)
+{
+   // Select the r'th available broker and use it to establish a tunnel
+   int i = -1;
+   int best = 0;
+   for (i = 0; i < broker_cnt; i++) {
+     if (brokers[i].ctx->standby_available &&
+         (brokers[i].ctx->usage < brokers[best].ctx->usage)) {
+       best = i;
+     }
+   }
+
+   brokers[best].ctx->standby_only = 0;
+   brokers[best].ctx->state = STATE_GET_COOKIE;
+   return best;
+}
+
 int broker_selector_first_available(broker_cfg *brokers, int broker_cnt, int ready_cnt)
 {
   // Select the first available broker and use it to establish a tunnel
@@ -1162,6 +1209,7 @@ void show_help(const char *app)
     "       -s hook       hook script\n"
     "       -t id         local tunnel id (default 1)\n"
     "       -L limit      request broker to set downstream bandwidth limit (in kbps)\n"
+    "       -a            select broker based on use\n"
     "       -g            select first available broker to connect to\n"
     "       -r            select a random broker\n"
   );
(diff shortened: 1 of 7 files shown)


mitar:
So how we are currently doing in wlan slovenija is that we connect to all VPN servers at the same time and then leave to the higher-level routing protocol to decide over which connection to route.

Can you provide a user case why would it be beneficial to connect to only one VPN server? Do you have issues with VPN server resources? Connection should not take so much resources?

lynxis:
@mitar is this new? I though you're only connecting the routers to **one** tunneldigger server. But the router choose the broker which respond first.

notice: i don't care about mtu here, it's another topic.

freifunk-berlin use a simple openvpn setup for the exit-nodes. These exit-nodes doesn't mesh over the vpn. Only provide internet connectivity via VPN. In Germany we got some legal bullshit which requires this in most cases.

When scaling upwards you will saturate a 100mbit port fast, because it only requires around 10-20 active internet users.
In Berlin we can even saturate 1000mbit ports of a VPN server, so we have to scale up to multiple servers. Let the client do the load balance in the best way, because it reduce the management effort.

lynxis:
updated define USAGE = 0x0a instead of 0x10.

mitar:
> is this new? I though you're only connecting the routers to one tunneldigger server. But the router choose the broker which respond first.

No, this is from the very beginning.

> These exit-nodes doesn't mesh over the vpn. Only provide internet connectivity via VPN. In Germany we got some legal bullshit which requires this in most cases.

So legal reasons prevent you from running a routing protocol on exists? Because if not, you could still simply connect all exists and then you use routing protocol to decide which exit to use.

> When scaling upwards you will saturate a 100mbit port fast, because it only requires around 10-20 active internet users.

But only one connection is effectively consumed. Because traffic is routed over it. Other connections are just established and ready to be used, but not much is send there (except for keep alive and other control packets and routing packets).

> In Berlin we can even saturate 1000mbit ports of a VPN server, so we have to scale up to multiple servers. Let the client do the load balance in the best way, because it reduce the management effort

I think that would be better done on the routing level. What do you use for routing? So router could know how much each VPN server link is used and route accordingly. For example, in the similar way how you can route based on the WiFi channel bandwidth use. This would even be better than a simple round robin because it could change after initial connection.

(BTW, I am not claiming that this feature is not OK, I am just asking about the rationale for it.)

broker/l2tp_broker.py
```diff
@@ -916,6 +918,22 @@ def issue_cookie(self, endpoint):
     self.cookies.put(endpoint, cookie)
     return cookie
 
+  def usage(self, endpoint):
+    """
+    returns usage information
+
+    :param endpoint: Endpoint tuple
+    :return: Usage information (1 byte)
+      0 (broker is not used)..255 (broker is used to hard)
```
mitar: `to` -> `too`

kostko:
> is this new? I though you're only connecting the routers to one tunneldigger server. But the router choose the broker which respond first.

We are using broker selection functionality just to select the best port for a broker, to bypass some firewall filtering. But then we use multiple tunnel interfaces to different brokers and a higher-layer routing protocol decides which one to use.

lynxis:
I see. But Berlin use olsr (v1) on the mesh. So for load-balancing I still prefer the usage, because we can only change the routing for all nodes at once, not just for 20 nodes go to that gateway, 20 to another gateway, etc.
Back to my original question:
> Is the usage a optional feature? Configurable? Or do we want all new servers supporting this, but still be compatible to old clients?

It would say I implement this as optional feature, ok?
When a client with this feature enabled ask an old broker, the client treats this broker as same as maximum clients, only using it as fallback.
A broker can give out the usage, but don't care if the client has requested usage.

mitar:
> But Berlin use olsr (v1) on the mesh.

We are using that with OLSRv1 as well. :-)

I will leave to @kostko to propose how this should be done. I do not see why we should not support sending load/stats information from the broker by default (maybe it should be a bit more future proof and allow usage to have both version of usage information and then payload? or should it also have some key=value for payload already?). And then clients can opt-in to respect this information and connect in specific order, or not.

lynxis:
key=value parser would involve a lot of code.
adding another byte for a version sound good.
The server can still decide on what metric it calculates it's usage. Is the metric just the clients connected or a bandwidth usage.

mitar:
About version, is there already a protocol-level version available? Maybe then we do not need also usage-level version.

kostko:
I agree that the simpler approach should be taken (1 or 2 bytes interpreted as an unsigned integer metric). Version byte is already part of the header, so unless the usage format will change more often than the protocol version, I don't think an additional version is necessary.

Compatibility with old clients/servers must be ensured in all combinations. In case one side doesn't support usage, it should be ignored and behave in the same way as it does currently.

lynxis:
rebase my patchset to ping travis.

lynxis:
rebased to fix conflicts.

mitar:
What is the status of this? Is this ready for review? I made a quick pass and I like it.

lcb01a:
This feature is also great for batman-advanced based mesh communities. We have already implemented tunneldigger as replacement for our user space VPN daemon and it is performing great.
We are using @lynxis patch on our brokers right now and it is working without any issues so far. The usage is evenly distributed on the two brokers. 
I've also tested the backwards compatibility and have not encountered any issues as well.

mitar:
Thank you for your report.

@kostko: how will this integrate with your rewrite?

kostko:
It looks like this is mostly a client-side change, so it should not be hard to integrate with the broker-rewrite branch (which will become master soon). @lynxis, any thoughts?

kostko:
I have written a quick patch against broker-rewrite, please see c026068c0b57e714ada4f8dbcc5b19bb8c04bb03 if it looks ok.

kostko:
Please rebase the pull request on master. Broker-side usage support is already there, so you can remove that part.

client/l2tp_client.c
```diff
@@ -963,6 +984,15 @@ void context_process(l2tp_context *ctx)
       }
       break;
     }
+    case STATE_GET_USAGE: {
+      // Get usage information if available
+      if (!is_timeout(&ctx->timer_usage, 2))
+        context_send_packet(ctx, CONTROL_TYPE_USAGE, "UUUUUUUU", 8);
+      else
+        /* time out. The broker might not support usage. ignore it. */
```
kostko: Can we use consistent comments (in this case `//`)? And start them with a capital letter.

client/l2tp_client.c
```diff
@@ -1268,24 +1353,20 @@ int main(int argc, char **argv)
...
         break;
-    }
 
-    // Select the first available broker and use it to establish a tunnel
-    main_context = NULL;
-    for (i = 0; i < broker_cnt; i++) {
-      if (brokers[i].ctx->standby_available) {
-        brokers[i].ctx->standby_only = 0;
-        main_context = brokers[i].ctx;
+      /* first available broker just use the first one available */
```
kostko: Can we use consistent comments (in this case `//`)? And start them with a capital letter.

client/l2tp_client.c
```diff
@@ -1093,6 +1123,60 @@ void context_free(l2tp_context *ctx)
...
+typedef struct {
+    char *address;
+    char *port;
+    l2tp_context *ctx;
+  } broker_cfg;
+#define MAX_BROKERS 10
+
+int broker_selector_usage(broker_cfg *brokers, int broker_cnt, int ready_cnt)
+{
+    // Select the r'th available broker and use it to establish a tunnel
```
kostko: This has some strange indentation?

client/l2tp_client.c
```diff
@@ -1093,6 +1123,60 @@ void context_free(l2tp_context *ctx)
   free(ctx);
 }
 
+// List of brokers
+typedef struct {
+    char *address;
+    char *port;
+    l2tp_context *ctx;
+  } broker_cfg;
```
kostko: This has some strange indentation?

client/l2tp_client.c
```diff
@@ -1093,6 +1123,60 @@ void context_free(l2tp_context *ctx)
...
+      brokers[i].ctx->state = STATE_GET_COOKIE;
+      return i;
+    }
+  }
+  return -1;
+}
+
+int broker_selector_random(broker_cfg *brokers, int broker_cnt, int ready_cnt)
+{
+    // Select the r'th available broker and use it to establish a tunnel
```
kostko: This has some strange indentation?

client/l2tp_client.c
```diff
@@ -492,6 +499,19 @@ void context_process_control_packet(l2tp_context *ctx)
 
   // Check packet type
   switch (type) {
+    case CONTROL_TYPE_USAGE: {
+      if (ctx->state == STATE_GET_USAGE) {
+          // broker usage information
```
kostko: This has some strange indentation? Also, comments should start with a capital letter.

kostko:
I've added some style-related comments and restarted the tests. BTW, could we add some kind of test that broker selection works properly?

lynxis:
I'm working on that already. An important change was   https://github.com/wlanslovenija/tunneldigger/pull/33 to execute commands and get output.
```
        """
        - we need to check if the client version is fresh enought to support usage. otherwise SKIP
        - setup server A with non-usage version
        - setup server B with usage version without any client
        - setup server C with usage version but one client

        - start client conf ABC and check if it's connecting to server B
        - start client conf AB and check if it's connecting to server B
        """
```

kostko:
That is amazing, thanks a lot!

mitar:
@lynxis: What is the status of this pull request?

lynxis:
https://github.com/lynxis/tunneldigger/tree/usage_test_4 shows the progress. I tested now the most stuff. It's missing a try this 10 times.

lynxis:
pushed everthing into here. but there are now merge conflicts. will look on this later

lynxis:
tests are failing because of an timeout. we have to shorten our tests or we've to move to jenkins.

lynxis:
i would like to merge this now.

kostko:
Can you go over the code review comments I posted above. After addressing those, this looks good to merge.

lynxis:
oh, overlooked those. fixing them...

lynxis:
@kostko done.

kostko:
Looks good to me! As soon as the test completes, I'll merge.

lynxis:
just for the record: those tests can be improved ;)

mitar:
About improving tests. I think a better way to do it would be to use [build matrix](https://docs.travis-ci.com/user/customizing-the-build/). So we could provide through environment variable commit hashes we want to check against in `.travis.yaml` file and then Travis will do it in parallel.

I do this with [package versions](https://github.com/wlanslovenija/datastream/blob/master/.travis.yml#L5). To test against multiple combinations of dependencies we want to support. Past and future.

If I understand you, this is the thing you were saying that it could be improved? Time tests are taking?

mitar:
@lynxis: I see more changes to the usage selection here: https://github.com/wlanslovenija/tunneldigger/compare/master...ffrl:master

Care to make a pull request so that we pull those improvements to this repository?

PR #27, opened 2015-11-29
client: change variable type to int for the return value of getopt()

lynxis:
On some platforms (e.g. ppc) this check will otherwise fail,
because the cast priority is different.


mitar:
It looks that this is POSIX anyway. Thanks.

PR #30, opened 2015-12-18
Tunneldigger broker rewrite

kostko:
I've been setting up the broker on an OpenWrt-based device and noticed that the broker code really needs a rewrite. This is it. It gets rid of most of the external dependencies and improves code organization and performance. One change is also that cookies are no longer stored in a cache but are instead generated using an HMAC-based construction, requiring no state. The configuration file format is the same, just that some options are currently ignored.

Some things are still not yet implemented:
* Hooks.
* Applying traffic limits.

Both should be easy to add.

The way the broker is executed is now also a little different, from the toplevel directory you execute:
```
python -m broker.main /etc/tunneldigger.conf
```


kostko:
Hm, this is strange, the tests seem to pass, but they shouldn't because hooks are not yet implemented.

kostko:
cc @lynxis

broker/broker.py
```diff
@@ -0,0 +1,212 @@
...
+        :param uuid: Unique tunnel identifier received from the remote host
+        :param remote_tunnel_id: Remotely assigned tunnel identifier
+        :return: True if a tunnel has been created, False otherwise
+        """
+
+        now = time.time()
+
+        # Rate limit creation of new tunnels to at most one every 10 seconds to prevent the
+        # broker from being overwhelmed with creating tunnels, especially on embedded devices.
+        if self.last_tunnel_created is not None and now - self.last_tunnel_created < 10:
```
kostko: This should probably be made a configuration option.

lynxis:
good questions, why it can not find any tests.
I like the idea, to rewrite the tunnelbroker. We should also think about the ability to support a control socket. We don't have to implement this today, just to hold the door open to add a control socket for the broker.

kostko:
> We should also think about the ability to support a control socket. We don't have to implement this today, just to hold the door open to add a control socket for the broker.

I agree and this is now easy to add. We should support a simple control protocol for getting the broker status etc.

mitar:
We also talked that IPv6 support would be good (so IPv6 for tunnels themselves, not inside the network). @lynxis, interested in implementing that? :-)

kostko:
Ah! @mitar broke the tests in 375a37b0d296520678382ccd0896d238f2ada0a8 with various renames.

mitar:
But they didn't fail?

kostko:
Yes, because you broke them in such a way that they were not found. Because you renamed the tests file and nose requires a specific filename pattern.

kostko:
I fixed this in this branch, so that the tests filename is explicit.

kostko:
I have now implemented support for hooks, so the tests against all older versions are now passing. The only remaining thing is support for honoring traffic limits (proper control packets are already parsed and ACKed, we just need to apply the tc rules).

kostko:
Traffic limits have now also been implemented.

If there are no objections, I will be merging this into master. I have been testing the new version on one of our Mikrotik-based VPN servers and it seems to be running much better than the old one.

PR #36, opened 2016-04-08
add wireshark lua dissector

lynxis:


mitar:
Cam you document this file a bit? How to use it? What it is? Line comments?

lynxis:
yes. I know. I'll add some documentation about how to use this file to the /docs/. I just want to publish it. because I written that thing 2 years ago.
a dissector parses a raw data network package. so when you sniff packages with wireshark, it shows you what kind of package this is. And it also forward the l2tp data packages to the l2tp parser, so you can even see all packges inside the tunnel.

mitar:
This works both for the control and data packets?

lynxis:
right.

mitar:
Hm, I worry a bit how to assure we keep this parser in sync with any changes to the protocol. But probably we should not be changing the protocol much anyway. :-)

kostko:
This should probably go into a separate directory?

lynxis:
into which one? /docs/

lynxis:
it's missing btw. the pmtu notify

kostko:
Yes, I think `docs` should be ok.

mitar:
Hm, can you still add documentation how to use it?

PR #38, opened 2016-04-17
Debian client

jaka:


mitar:
Maybe this should be inside `client` directory? Because is packaging for `client`? Also `files` name is a bit strange.

But otherwise really cool. Maybe @kostko will have more comments.

jaka:
I am aware `files` name is not the most appropriate word and it should be moved under `client` directory. Then `debian/rules` should be updated as well. Moreover, I forgot to remove _git_ dependency from `debian/control` since it is not needed for building the binary anymore.

I think it should be cool to enhance building process so that a filename of a builded deb includes git revision.

lynxis:
@kostko @mitar what about creating a signed version tag? This would make it easier to maintain for jaka. But  I don't think we're ready for a release, there are still bugs in the broker atm.

mitar:
Yes, we should release a version at some point. But for now it would be great to have code for making a package in the repository.

> But I don't think we're ready for a release, there are still bugs in the broker atm.

Care to specify which you have in mind?

lynxis:
@jaka can you add a test for travis to produce an debian package?

lrnzo:
i cloned this and merged it into origin/v0.2.0
build on armhf worked
i pushed it all to https://git.nordwest.freifunk.net/lrnzo/tunneldigger.git

mitar:
Does it work with master branch?

RalfJung:
What is the status of this PR?

kaechele:
This would need to be reworked if #69 is merged.

RalfJung:
#69 got merged so this is outdated now. Closing due to inactivity.

@jaka feel free to re-open the PR if you want to give this another try, updated to the build system changes.

PR #41, opened 2016-05-02
client: remove #define HAVE_* and config.h

lynxis:
HAVE_* should set by a configure/cmake, not by the source.
config.h usually contains those variables.
This fixes build errors with openwrt trunk, because warnings treated as
Errors.


lynxis:
^another commit...

kostko:
Compile test seems to be failing.

lynxis:
@kostko it needs to be compil tested on OpenWrt/LEDE

PR #43, opened 2016-05-11
Introduce new states

lynxis:
based on branch small misc. Please merge #42 first.

### DIFF
--- a/client/asyncns.c
+++ b/client/asyncns.c
@@ -79,7 +79,7 @@ enum {
 };
 
 struct asyncns {
-    int fds[MESSAGE_FD_MAX];
+    int fds[4];
 
 #ifndef HAVE_PTHREAD
     pid_t workers[MAX_WORKERS];
@@ -196,8 +196,6 @@ static int close_allv(const int except_fds[]) {
 
     assert(except_fds);
 
-    /* We ignore FD_CLOEXEC here, since this is called in the child only anyway */
-
     if ((d = opendir("/proc/self/fd"))) {
 
         struct dirent *de;
@@ -272,7 +270,7 @@ static int close_allv(const int except_fds[]) {
         for (i = 0; except_fds[i] >= 0; i++)
             if (except_fds[i] == fd) {
                 found = 1;
-                break;
+                continue;
             }
 
         if (found)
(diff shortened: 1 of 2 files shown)


client/asyncns.c
```diff
@@ -77,7 +77,7 @@ enum {
 };
 
 struct asyncns {
-    int fds[MESSAGE_FD_MAX];
+    int fds[4];
```
mitar: Please use verbose-named constants instead of just numbers.

mitar:
#42 has been merged.

client/asyncns.c
lynxis: this is part of a syncronisation of the upstream project of this file.

mitar: Oh, this is 3rd party code? Why it is not in some `vendor` subdirectory or something. :-)

mitar: Have you submitted `#define _GNU_SOURCE` upstream as well?

lynxis: No, because they are using autoconf to define their config.h and a lot of defines.

mitar: Hm, but if we update this code next time from upstream, how can we know what to reapply? Maybe we should have some patch and apply it during our compilation? Or we should try to get those things upstream.

lynxis:
ToDos before merging:
[] fix comments (/* */ by //)
[] drop state ERROR

lynxis:
ready for merge.

mitar:
Could you comment somewhere (maybe the best in the commit message) which revision of upstream `asyncns.c ` has this now been? Because then later on we can easier make a diff against it, see what are our changes and apply them.

Even better it is if we have commit with original revision on its own in the history. Then when you want to update the upstream, you go to that commit again, you update the version, commit new (you are in detached head mode then), and then merge back into the latest of our repository. In this way then git then handles all our merging of our changes on top.

lynxis:
ping @kostko @mitar

client/l2tp_client.c
```diff
@@ -1097,6 +1151,8 @@ void context_process(l2tp_context *ctx)
       }
       break;
     }
+    case STATE_IDLE:
```
kostko: Please add braces as it is in other `case` statements.

kostko:
I see some fixes in the external asyncns. Are we testing DNS resolution?

kostko:
Apart from the minor comment above, this looks good to me. It would be very nice if we could also have tests for DNS resolution, but this is something that can be added in another PR.

mitar:
I think we are testing DNS resolution in 8bbc5e4a226f73aa990ff9a08c0441ac38bea4b8?

PR #44, opened 2016-05-11
tests: ping a fqdn instead of an ip when checking connectivity

lynxis:


lynxis:
strange. it doesn't fix the problem. but it can definitly resolve it. Do they block ubuntu/debian repositorys?

lynxis:
I still think it's a good idea, even it doesn't fix the whole problem. maybe travis is blocking access to the mirror server when executing the *script-section*

tests/tunneldigger.py
```diff
@@ -33,7 +33,7 @@ def setup_template():
             raise RuntimeError("failed to start container")
 
     container.attach_wait(lxc.attach_run_command, ["dhclient", "eth0"])
-    check_ping(container, '8.8.8.8', 10)
+    check_ping(container, 'ccc.de', 10)
```
mitar: Could we ping instead `google-public-dns-a.google.com`? I believe it might have higher uptime than `ccc.de`. ;-)

lynxis:
but if ccc.de is down, it's simpler to validate that, because 8.8.8.8 is an anycast address.

mitar:
But we are not testing here if `ccc.de` is up or down. We need it to be up always, no? I think we should not have test fail and then have to go and see that it failed because `ccc.de` was down. I think it is better that test never fails for this reason. And using Google anycast will assure that.

lynxis:
ok. i'll use the anycast address

lynxis:
@mitar pong

mitar:
Maybe we should then ping Travis CI servers? :-)

PR #47, opened 2016-05-13
Update client.rst

txt-file:
I needed the file netlink.h which is provided by linux-headers in debian.

### DIFF
--- a/docs/client.rst
+++ b/docs/client.rst
@@ -13,6 +13,11 @@ This will give you a ``tunneldigger`` directory which contains the broker and
 the client in separate directories. Client code can be compiled into a
 stand-alone program.
 
+Requirements
+------------
+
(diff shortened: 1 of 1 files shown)


mitar:
There is a pull request for Debian package: #38

RalfJung:
Compiling the client on Debian oldstable (jessie) and newer fails because it needs libnl1, which was dropped from the archive. The `netlink.h` in the kernel sources is (AFAIK) a totally different header; tuneldigger needs `netlink/netlink.h` while the kernel has `netfilter/netlink.h`.

So, I think this change is wrong.

kaechele:
Yup, this is factually incorrect.

RalfJung:
@kostko can you close this MR?

PR #48, opened 2016-06-13
Fix for more than 2 tunneldigger brokers:

Fungur69:
select() moved from context_process() to a new function do_select(),
so that packets get waited for for all brokers at once. After that all
received packages get processed. Previously packages from brokers would
not be processed while waiting for packages for other brokers, which
could run into the 1 sec timeout.
Sideeffect: some superfluous control packets will be sent in the
broker selection phase.

### DIFF
--- a/client/l2tp_client.c
+++ b/client/l2tp_client.c
@@ -117,7 +117,9 @@ enum l2tp_limit_type {
 };
 
 enum l2tp_ctrl_state {
-  STATE_GET_USAGE,
+  STATE_IDLE, // doing nothing
+  STATE_USAGE_SEND, // send out a usage package
+  STATE_USAGE_WAIT, // wait for receiption
   STATE_GET_COOKIE,
   STATE_GET_TUNNEL,
   STATE_KEEPALIVE,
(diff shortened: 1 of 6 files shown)


kostko:
Tests seem to be failing due to a compile error in line 250, where you reference `ctx`, which is not defined:
```
FD_SET(ctx->fd, &rfds);
```

This line should just be removed as you configure fds later for all brokers.

client/l2tp_client.c
```diff
@@ -205,6 +205,14 @@ typedef struct {
   time_t timer_resolving;
 } l2tp_context;
 
+// List of brokers
+typedef struct {
+    char *address;
```
kostko: Fix indentation.

kostko:
Something is strange with your last commit and Travis is getting confused for some reason and checking out a different commit, which is not actually there. The history shows "commited on GitHub", so this could be the reason.

Can you rebase your commits and push over here again?

mitar:
This branch has conflicts anyway.

Fungur69:
I merged the latest changes. I hope this builds now.

mitar:
Tests are passing!

PR #49, opened 2017-04-01
New CFFI-based conntrack interface

kostko:
This PR introduces the following changes:
* The broker is packaged as a Python package, installed via `setup.py` as `tunneldigger_broker`. It can then be invoked by running `python -m tunneldigger_broker.main <cfg>`.
* The old ctypes-based conntrack interface is replaced with a newer CFFI-based one. This means that building the broker package now requires a working compiler. A major upside is that the code now depends on the conntrack API (will cause build-time failure) instead of the conntrack ABI (will cause run-time segmentation faults). Another point is that this uses a different method of passing callbacks, which should hopefully fix a crash on LEDE x86_64.


mitar:
Looks good.

PR #50, opened 2017-10-29
Fix checking for netlink errors when receiving only a single message

RalfJung:
The current implementation fails to raise an exception when an error is returned.

This masks a problem where on newer kernels, tunneldigger fails to create tunnels because l2tpv3 session IDs have to be globally unique.

### DIFF
--- a/broker/src/tunneldigger_broker/netlink.py
+++ b/broker/src/tunneldigger_broker/netlink.py
@@ -153,6 +153,17 @@ class Connection:
         self.descriptor.send(msg)
 
     def recv(self, multiple = False):
+        """
+        Receive message(s) from the netlink socket.
+
+        :param multiple: Control the receiving mode.
+
+        In single-message mode, exactly one message is received.  If that message is an error,
+        an exception is raised.  Otherwise (including if the message is an ack), it is returned.
+        In multiple-message mode, messages are received until the first NLMSG_DONE.
+        If there is an error, an exception is raised.  Acks and the final NLMSG_DONE are dropped.
+        The other messages are returned in a list.
+        """
         messages = []
         done = False
 
--- a/broker/src/tunneldigger_broker/netlink.py
+++ b/broker/src/tunneldigger_broker/netlink.py
@@ -165,6 +165,14 @@ class Connection:
                 msg.pid = pid
                 contents = contents[msglen:]
 
+                if msg.type == NLMSG_ERROR:
+                    errno = -struct.unpack("i", msg.payload[:4])[0]
+                    if errno != 0:
+                        err = OSError("Netlink error: %s (%d)" % (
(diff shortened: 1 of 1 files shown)


kostko:

RalfJung:
Done.

broker/src/tunneldigger_broker/netlink.py
```diff
@@ -165,6 +165,14 @@ def recv(self, multiple = False):
...
                 contents = contents[msglen:]
 
+                if msg.type == NLMSG_ERROR:
+                    errno = -struct.unpack("i", msg.payload[:4])[0]
+                    if errno != 0:
+                        err = OSError("Netlink error: %s (%d)" % (
+                                                             os.strerror(errno), errno))
+                        err.errno = errno
+                        raise err
+                # A non-error message
```
kostko: Add an empty line above.

broker/src/tunneldigger_broker/netlink.py
```diff
@@ -173,12 +181,8 @@ def recv(self, multiple = False):
                     done = True
                     break
                 elif msg.type == NLMSG_ERROR:
-                    errno = -struct.unpack("i", msg.payload[:4])[0]
-                    if errno != 0:
-                        err = OSError("Netlink error: %s (%d)" % (
-                                                             os.strerror(errno), errno))
-                        err.errno = errno
-                        raise err
+                    # an ack
```
kostko: This is only possible in case the `errno` contained in the message is zero. Currently, the handling in this case is inconsistent when `multiple` is `True` (error message is skipped and not included in the final list) and when `multiple` is `False` (error message is appended to the list of received messages). I think this should be unified.

RalfJung: When `multiple` is `False`, the entire point of calling this message is to receive the ACK. If we skip it, all uses of `multiple=False` will break.

Maybe it makes more sense to have a dedicated function `recv_ack`, and rename this one into `recv_until_done`?

kostko: I see, so an ACK is actually a message of type `NLMSG_ERROR`, but with an errno payload of zero. The problem with two methods is that they would most likely be very similar, so perhaps it would be better to just document this behavior in the method's documentation block.

This whole netlink code is really ugly as it was imported from somewhere else and we should eventually refactor it.

RalfJung: > an ACK is actually a message of type NLMSG_ERROR, but with an errno payload of zero.

Correct.

> better to just document this behavior in the method's documentation block.

Will do.

RalfJung:
I rebased and added documentation to `recv`.

RalfJung:
Rebased.

PR #51, opened 2017-10-29
Implement some missing config options

RalfJung:
There is a significant discrepancy between the config options listed in l2tp_broker.cfg.example and the ones actually supported.  This PR fixes that by, mostly, implementing `log.verbosity` and `log.log_ip_addresses`. Furthermore, it makes the hard-coded limit of one connection per 10s configurable.

### DIFF
--- a/broker/src/tunneldigger_broker/hooks.py
+++ b/broker/src/tunneldigger_broker/hooks.py
@@ -96,7 +96,7 @@ class HookManager(object):
     Manages hooks.
     """
 
-    def __init__(self, event_loop):
+    def __init__(self, event_loop, log_arguments):
         """
         Constructs a new hook manager instance.
 
@@ -106,6 +106,7 @@ class HookManager(object):
         self.event_loop = event_loop
         self.hooks = {}
         self.processes = {}
+        self.log_arguments = log_arguments
 
         # Create a file descriptor so we can get notified of SIGCHLD signals in the
         # context of the event loop (and not in an arbitrary location).
@@ -141,7 +142,10 @@ class HookManager(object):
         if not script:
             return
 
-        logger.info("Running hook '%s' via script '%s %s'." % (name, script, " ".join([str(x) for x in args])))
+        if self.log_arguments:
+            logger.info("Running hook '%s' via script '%s %s'." % (name, script, " ".join([str(x) for x in args])))
+        else:
+            logger.info("Running hook '%s' via script '%s'." % (name, script))
         try:
             process = HookProcess(name, script, args)
             process.register(self.event_loop)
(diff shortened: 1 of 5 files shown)


RalfJung:
I hope I found all places where IP addresses are being logged; please let me know if I did not. It is important for us that we do not log users' IP addresses, under any circumstances.

kostko:

RalfJung:
Done.

broker/src/tunneldigger_broker/hooks.py
```diff
@@ -141,7 +142,10 @@ def run_hook(self, name, *args):
         if not script:
             return
 
-        logger.info("Running hook '%s' via script '%s %s'." % (name, script, " ".join([str(x) for x in args])))
+        if self.log_arguments:
+            logger.info("Running hook '%s' via script '%s %s'." % (name, script, " ".join([str(x) for x in args])))
+        else:
+            logger.info("Running hook '%s' via script '%s'." % (name, script))
```
kostko: Please add an empty line below.

broker/src/tunneldigger_broker/main.py
```diff
@@ -55,7 +55,7 @@
 event_loop = eventloop.EventLoop()
 
 # Initialize the hook manager.
-hook_manager = hooks.HookManager(event_loop)
+hook_manager = hooks.HookManager(event_loop, log_arguments=config.getboolean('log', 'log_ip_addresses'))
```
kostko: Break into multiple lines.

RalfJung:
I fixed both formatting issues.

kostko:
Great, just rebase on master and it should be good to merge.

RalfJung:
> rebase on master

Again? Oh well.^^ Done.

kostko:
Yes, you need to do it everytime something gets merged into master ;-) I can also do it for you, but then the commits will get rewritten.

RalfJung:
> Yes, you need to do it everytime something new gets merged into master ;-)

Why that? We don't do that other projects I have contributed to. We rather preserve the actual (non-linear) history of development.

kostko:
I prefer having a linear history as merge commits can mask what was actually modified (e.g. in case there are conflicts, which require code changes and those changes are then made in the merge commit).

RalfJung:
> I prefer having a linear history as merge commits can mask what was actually modified (e.g. in case there are conflicts, which require code changes and those changes are then made in the merge commit).

Personally, I'd rather have the fact that there were conflicts documented by having a merge commit, than doing rebases which entirely erases such conflicts from living memory.

But well, your repository, your rules. I can live with that :)

PR #52, opened 2017-10-29
update broker documentation

RalfJung:
The Fedora package names are still wrong, I don't know Fedora well enough to update them.


RalfJung:
CI fails after just changing documentation? I doubt that is caused by this PR, then.

RalfJung:
Rebased, let's see if this works now.

PR #53, opened 2017-10-29
Fix computing the set of available tunnel_ids

RalfJung:
With a tunnel_id_base of 100 and max_tunnels 100, the valid tunnel_ids range from 100 inclusive to 200 exclusive. So we want to call `xrange(100, 200)`. No need to add `1` more, that will actually add 101 tunnel IDs.

Patch by @rohammer

### DIFF
--- a/broker/src/tunneldigger_broker/broker.py
+++ b/broker/src/tunneldigger_broker/broker.py
@@ -32,7 +32,7 @@ class TunnelManager(object):
         self.hook_manager = hook_manager
         self.max_tunnels = max_tunnels
         self.tunnel_id_base = tunnel_id_base
-        self.tunnel_ids = set(xrange(tunnel_id_base, tunnel_id_base + max_tunnels + 1))
+        self.tunnel_ids = set(xrange(tunnel_id_base, tunnel_id_base + max_tunnels))
         self.tunnel_port_base = tunnel_port_base
         self.namespace = namespace
         self.tunnels = {}


kostko:

RalfJung:
Done.

broker/src/tunneldigger_broker/broker.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, hook_manager, max_tunnels, tunnel_id_base, tunnel_port_base,
         self.hook_manager = hook_manager
         self.max_tunnels = max_tunnels
         self.tunnel_id_base = tunnel_id_base
-        self.tunnel_ids = set(xrange(tunnel_id_base, tunnel_id_base + max_tunnels + 1))
+        self.tunnel_ids = set(xrange(tunnel_id_base, tunnel_id_base + max_tunnels))
```
kostko: While we are changing this, can we switch to `range` instead?

RalfJung: But `xrange` is better because it doesn't allocate a list, right?

kostko: Yes, it just returns a generator. But in Python 3 it got removed as `range` returns a generator there. But perhaps we should just leave it for now and handle it in a later "Python 3 support" PR.

RalfJung: I think it'd make more sense to just at some point (the sooner the better :D ) move to Python 3 and drop support for Python 2.

But, if you want, I can also do `s/xrange/range/` here and now.

RalfJung:
Would you prefer me to just rebase or to rebase + `s/xrange/range/`?

kostko:
Merged in feb4f15aab6e9ecd58d95e4e208f762e0b70a981.

kostko:
Oh sorry, didn't see your comment, I already rebased and merged into master.

RalfJung:
That's fine, less work for me and it does preserve the original commit author :)

Issue #55, opened 2017-10-30
Broker crashes, running out of file descriptors


RalfJung:
After 5-6h of uptime, the tunneldigger broker quits with the following error:

```
Oct 30 03:24:41 gw3.saar.freifunk.net python[9594]: Traceback (most recent call last):
Oct 30 03:24:41 gw3.saar.freifunk.net python[9594]: File "/usr/lib/python2.7/runpy.py", line 162, in _run_module_as_main
Oct 30 03:24:41 gw3.saar.freifunk.net python[9594]: "__main__", fname, loader, pkg_name)
Oct 30 03:24:41 gw3.saar.freifunk.net python[9594]: File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
Oct 30 03:24:41 gw3.saar.freifunk.net python[9594]: exec code in run_globals
Oct 30 03:24:41 gw3.saar.freifunk.net python[9594]: File "/opt/tunneldigger/lib/python2.7/site-packages/tunneldigger_broker-0.3.0-py2.7-linux-x86_64.egg/tunneldigger_broker/main.py", line 113, in <module>
Oct 30 03:24:41 gw3.saar.freifunk.net python[9594]: event_loop.start()
Oct 30 03:24:41 gw3.saar.freifunk.net python[9594]: File "/opt/tunneldigger/local/lib/python2.7/site-packages/tunneldigger_broker-0.3.0-py2.7-linux-x86_64.egg/tunneldigger_broker/eventloop.py", line 59, in start
Oct 30 03:24:41 gw3.saar.freifunk.net python[9594]: pollable.read(file_object)
Oct 30 03:24:41 gw3.saar.freifunk.net python[9594]: File "/opt/tunneldigger/local/lib/python2.7/site-packages/tunneldigger_broker-0.3.0-py2.7-linux-x86_64.egg/tunneldigger_broker/network.py", line 98, in read
Oct 30 03:24:41 gw3.saar.freifunk.net python[9594]: callback()
Oct 30 03:24:41 gw3.saar.freifunk.net python[9594]: File "/opt/tunneldigger/local/lib/python2.7/site-packages/tunneldigger_broker-0.3.0-py2.7-linux-x86_64.egg/tunneldigger_broker/tunnel.py", line 230, in pmtu_di
Oct 30 03:24:41 gw3.saar.freifunk.net python[9594]: self.create_timer(self.pmtu_discovery, timeout=random.randrange(2, 5))
[...]

PR #59, opened 2017-10-30
close hook process file descriptors

RalfJung:
This helps with #55. Now I am no longer seeing hundreds of open pipe file descriptors.

There are still many many timerfd though, so I am wondering if there is another leak.

### DIFF
--- a/broker/src/tunneldigger_broker/hooks.py
+++ b/broker/src/tunneldigger_broker/hooks.py
@@ -72,6 +72,8 @@ class HookProcess(object):
             pass
 
         self.process.poll()
+        self.process.stdout.close()
+        self.process.stderr.close()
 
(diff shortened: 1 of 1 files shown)


kostko:
> There are still many many timerfd though

How many? They should be proportional to the number of open tunnels as each tunnel can have a few timers.

RalfJung:
It's hard to tell yet whether it is proportional or not. One server has:
```
# ls -lah | fgrep socket -c
77
# ls -lah | fgrep timer -c
146
```
and the other one
```
# ls -lah | fgrep socket -c
102
# ls -lah | fgrep timer -c
196
```

I suppose after all the nightly reconnects, tomorrow we will see whether this stays at roughly factor 2.

PR #60, opened 2017-10-31
explain some config options better

RalfJung:


RalfJung:
I rebased onto master.

PR #61, opened 2017-10-31
Avoid reconnecting to a broker that failed on us

RalfJung:
Currently, `broker_selector_usage` returns `0` when no broker is ready. That's just wrong, a selector should never return a broker that is not ready. It should return `-1` instead, like the others, which will make broker selection restart.

EDIT: I expanded the scope of the PR; see below.

### DIFF
--- a/client/l2tp_client.c
+++ b/client/l2tp_client.c
@@ -229,12 +229,12 @@ static asyncns_t *asyncns_context = NULL;
 
 int broker_selector_usage(broker_cfg *brokers, int broker_cnt, int ready_cnt)
 {
-   // Select the r'th available broker and use it to establish a tunnel.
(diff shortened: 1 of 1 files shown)


RalfJung:
While I was at it, I also improved the logging of the broker usage.

RalfJung:
I expanded the scope of the PR to also include a new feature: If connecting to a broker fails, we will not try that broker again because we assume it is broken. This only applies to failure during the initial connection establishment; if the broker disconnects later (e.g. because it got restarted), that's fine. When all brokers got disabled, we enable them all again.

This avoids silly loops where for some reason, one broker doesn't work, and hence it has the lowest usage, and hence everyone picks it, and hence nobody can connect. It doesn't catch more higher-level failure though (e.g. l2tp works, but then there is no batman connection).

Issue #62, opened 2017-11-01
MTU auto-detection does not work properly


RalfJung:
Ever since we upgraded two of our machines from the old (pre-rewrite) tunneldigger to current master, we have MTU issues on those servers. With the old setup, we used to disable the automatic MTU discovery, setting the MUT to a pretty conservative (low) value of 1406 manually.

I have no experience debugging MTU issues, so for now, I am going to re-implement the option to disable MTU discovery. But that's not solving this issue.

---

In case that matters, here are some more details of our setup: We are adding all the tunnels to a single bridge. We set that bridge's MTU to 1406 after it is created; however, as far as I understand, that value does not have any effect once a device is added to the bridge.

I am aware that to use the MTU properly, we would need one bridge per MTU. However, putting them all into the same bridge will use the minimum MTU, so things should still work, right?

kostko:
Are you sure that the PMTU is discovered incorrectly? Can you test if it works, e.g. by using ICMP ECHO packets of the given sizes? Can you give an example of a correct PMTU and the PMTU detected by tunneldigger?

> I am aware that to use the MTU properly, we would need one bridge per MTU. However, putting them all into the same bridge will use the minimum MTU, so things should still work, right?

Yes, we have always used such a configuration (one bridge per MTU). 

Putting all of them into the same bridge should use the minimum MTU on the transmit side. But the problem is that you actually have two MTUs (on the transmit and receive sides) and they can be different. In case you use a bridge, any traffic that gets sent into the bridge interface will indeed honor the MTU of the bridge (minimum of all bridge ports). But a bridge port may still receive a packet with a higher MTU from a port which supports such a higher MTU, which will (AFAIK) cause the packet to be silently dropped. So for best results, you must ensure that the MTU on both ends of the tunnel are consistent otherwise you could have issues.

RalfJung:
> Can you give an example of a correct PMTU and the PMTU detected by tunneldigger?

Unfortunately not. I have no idea what the correct PMTUs are. But I know that applying #63 fixed our server from "unusably slow / no connection" to "working properly".

> Can you test if it works, e.g. by using ICMP ECHO packets of the given sizes?

[...]

PR #63, opened 2017-11-01
provide an option to set a fixed MTU

RalfJung:
Provides a work-around for #62


RalfJung:
Thanks :)

I will still try to debug #62 though. Maybe this is a good chance for me to actually understand some of this MTU mess...

PR #64, opened 2017-11-09
Unique session ids

### LINKED ISSUE #57: Recent kernels enforce system-wide unique L2TPv3 session IDs
When running tunneldigger on the current Debian stable kernel (4.9.51), only one client can connect. The second client fails because the l2tp tunnel interface does not appear. After fixing a bug in the netlink interface (#50), one can see that the kernel sends an EEXIST in reply to the session_create.

A lot of digging through the linux kernel sources uncovered the source of the issue: L2TPv3 session IDs have to be unique system-wide. Tunneldigger hard-codes a session ID of 1 for every connection. That used to work due to a bug in the kernel, which meant that the kernel failed to actually ensure uniqueness of the session ID. That bug got fixed by <https://github.com/linux-stable/linux-stable/commit/dbdbc73b44782e22b3b4b6e8b51e7a3d245f3086>, which was backported to a few stable series, in particular, to 4.9.36.

## Proposed fix

Fixing this in a compatible way will require protocol changes: Both ends of the tunnel have to know each others session ID, so they have to negotiate whether they use 1 or something more unique. I started working on a fix at <https://github.com/freifunk-saar/tunneldigger/tree/wlanslovenija>. The approach is summarized in the commit message over there, copied here for reference:

```
This patch adds unique session IDs to tunneldigger in a backwards-compatible
way.  If both ends of the tunnel agree to use a unique session ID, they both
will use the tunnel ID as the session ID.  To manage this mutual agreement, two
messages in the protocol are changed:

CONTROL_TYPE_PREPARE gains a new optional byte at the end that clients use to
indicate to the server whether they want to use a unique session ID.  Old
servers will just ignore this additional byte.  New servers now know they are
talking with a modern client, and use unique session IDs for this connection.
New servers talking with old clients will notice the absence of this request and
use 1 as the session ID.

Furthermore, CONTROL_TYPE_TUNNEL gains a new optional byte at the end that
servers use to tell clients that they acknowledge using unique session IDs.  Old
clients will never see this additional byte, as the server only sends it if
unique session IDs were requested in CONTROL_TYPE_PREPARE.  New clients know,
upon seeing this byte, that they are talking to a new server, and will hence use
unique session IDs.  If a new client talks to an old server, it will receive an
old-style CONTROL_TYPE_TUNNEL and hence know that it has to use session ID 1.

So, both old a new clients can talk with both old and new servers.  However, of
course, if the server has a recent enough kernel, even though it can communicate
with old clients, it still can only support one old client at a time.
```

I am running two of our four servers with this fix, so compatibility with old clients is tested already. However, due to #55, I can't say anything about long-time stability yet. I also couldn't yet test new clients as I am still fighting my firmware build system. (The client uses such an ancient version of libnl that I can't build it on the host.)

## Open problems

As the last paragraph in the commit description says, there still is a potential problem: Once we upgrade one of our servers to a kernel including the problematic bugfix, only one old client will be able to talk to it at a time. There is nothing we can do about this, but what want to avoid is a client trying to connect to a new server and failing, while there are old servers (with higher usage) that could still support this client. I first tried to (ab)use `CONTROL_TYPE_USAGE` to let the client indicate whether it supports unique session IDs, so that the server could report "I am full" to old clients and steer them elsewhere. However, clients actually seem to send some rather arbitrary data alongside that message (`UUUUUUUU`, to be precise -- wtf?!?), so I am worried that attaching meaningful bytes here will not work very well. We could introduce a `CONTROL_TYPE_USAGE2`, but I think I have a better idea.

Clients already have a retry loop to connect again if the connection to the broker failed. I think clients should remember which broker failed, and exclude that one in the next round. Only once all brokers got excluded that way, they will be enabled again. This will, I think, improve client behavior in general, not just for this particular issue. It will also solve this issue as (after #50), brokers will send an error to clients when the session ID is already used, making the client try some other server. So, as long as one of the available brokers still has an old kernel, old clients will reliably be able to connect. Furthermore, even if all servers are on a new kernel, there can still be N old clients connected at the same time (and hopefully, they will fetch an auto-update and then become new clients).

I started implementing this, but got stuck yesterday due to the aforementioned build system issues.

RalfJung:
Quoting the main commit message:

```
    L2TPv3 session IDs have to be unique on the entire system, not just per tunnel.
    The only reason that tunneldigger got away with using 1 for all sessions is that
    older Linux kernels failed to properly check for duplicate session IDs.  That
    got fixed by kernel commit dbdbc73b44782e22b3b4b6e8b51e7a3d245f3086.
    
    This patch adds unique session IDs to tunneldigger in a backwards-compatible
    way.  If both ends of the tunnel agree to use a unique session ID, they both
    will use the tunnel ID as the session ID.  To manage this mutual agreement, we
    introduce space for up to 32 feature flags.  Three messages are affected:
    
    CONTROL_TYPE_USAGE optionally takes 4 additional bytes after the padding, used
    to send the client's feature flags.  Brokers may skew the usage they report back
    depending on client features.  Old clients will not send these bytes, which is
    interpreted as all flags being 0.
    
    CONTROL_TYPE_PREPARE is extended the same way, also sending the client's feature
    flags, so that the server does not have to remember them.  These are the
    features that the client is offering to be using for this session.  Old clients
    will not send any flags, which tells the server that no features are supported.
    Old servers will ignore flags if they received them.
    
    CONTROL_TYPE_TUNNEL is extended the same way, reporting the features that are
    actually going to be used for this session.  Typically, this will be the
    client's features masked by what the server supports.  If a feature-aware client
    talks to an old server, the flags are going to be missing, telling the client
    that no features are going to be used in this session.
    
    The first feature flag is 'unique session IDs'.  If enabled, the session ID of
    each side of the connection is going to be the same as the tunnel ID.
    Furthermore, the broker gains an option to report full usage for clients not
    supporting unique session IDs, making sure they connect to other servers (if any
    are available).
```

I think this is ready to be reviewed. However, before merging, I'd like to
* [x] roll out an experimental firmware showing that the client stuff really works, and 
* [x] at least try to figure out why we still have two clients connecting to the Gateway that now should report usage 0xFFFF.

Fixes #57

### DIFF
--- a/broker/src/tunneldigger_broker/protocol.py
+++ b/broker/src/tunneldigger_broker/protocol.py
@@ -37,6 +37,10 @@ ERROR_REASON_TIMEOUT        = 0x03
 ERROR_REASON_FAILURE        = 0x04 # e.q. on malloc() failure
 ERROR_REASON_UNDEFINED      = 0x05
 
+# Session feature flags
+FEATURE_UNIQUE_SESSION_ID = 1 << 0
+FEATURES_MASK = FEATURE_UNIQUE_SESSION_ID
+
 # Limit types.
 LIMIT_TYPE_BANDWIDTH_DOWN = 0x01
 
@@ -123,14 +127,14 @@ class HandshakeProtocolMixin(object):
             offset = 0
 
             # Verify cookie value.
-            timestamp = msg_data[offset:offset+2]
+            timestamp = msg_data[offset:offset + 2]
             offset += 2
             signed_value = '%s%s%s' % (address[0], address[1], timestamp)
             signature = hmac.HMAC(SECRET_KEY, signed_value, hashlib.sha1).digest()[:6]
             timestamp = struct.unpack('!H', timestamp)[0]
 
             # Reject message if more than 2 protocol ticks old.  One tick is 1 >> 6 = 64 seconds.
-            if signature != msg_data[offset:offset+6] or abs(protocol_time() - timestamp) > 2:
+            if signature != msg_data[offset:offset + 6] or abs(protocol_time() - timestamp) > 2:
                 return
             offset += 6
 
@@ -164,7 +168,7 @@ class HandshakeProtocolMixin(object):
 
             client_features = 0
             try:
-                client_features = struct.unpack('!I', msg_data[8:8+4])[0]
+                client_features = struct.unpack('!I', msg_data[8:8 + 4])[0]
             except (struct.error, IndexError):
                 pass
 
--- a/broker/src/tunneldigger_broker/protocol.py
+++ b/broker/src/tunneldigger_broker/protocol.py
@@ -112,23 +112,45 @@ class HandshakeProtocolMixin(object):
             signature = hmac.HMAC(SECRET_KEY, signed_value, hashlib.sha1).digest()[:6]
             self.write_message(address, CONTROL_TYPE_COOKIE, timestamp + signature)
         elif msg_type == CONTROL_TYPE_PREPARE:
+            # Packet format:
+            # 2 bytes protocol time
+            # 6 bytes "cookie", must be the signature we computed above
+            # 1 byte UUID length
+            # N bytes UUID
+            # optionally 4 bytes remote tunnel ID
+            # optionally 4 bytes client feature flags
+
+            offset = 0
+
             # Verify cookie value.
-            timestamp = msg_data[:2]
+            timestamp = msg_data[offset:offset+2]
+            offset += 2
             signed_value = '%s%s%s' % (address[0], address[1], timestamp)
             signature = hmac.HMAC(SECRET_KEY, signed_value, hashlib.sha1).digest()[:6]
             timestamp = struct.unpack('!H', timestamp)[0]
 
-            if signature != msg_data[2:8] or abs(protocol_time() - timestamp) > 2:
+            # Reject message if more than 2 protocol ticks old.  One tick is 1 >> 6 = 64 seconds.
+            if signature != msg_data[offset:offset+6] or abs(protocol_time() - timestamp) > 2:
                 return
+            offset += 6
 
-            uuid_len = struct.unpack('!B', msg_data[8])[0]
-            uuid = msg_data[9:9 + uuid_len]
+            uuid_len = struct.unpack('!B', msg_data[offset])[0]
+            offset += 1
+            uuid = msg_data[offset:offset + uuid_len]
+            offset += uuid_len
+
+            # Parse optional data
+            remote_tunnel_id = 1
+            client_features = 0
             try:
-                remote_tunnel_id = struct.unpack('!I', msg_data[10 + uuid_len:10 + uuid_len + 4])[0]
-            except struct.error:
-                remote_tunnel_id = 1
+                remote_tunnel_id = struct.unpack('!I', msg_data[offset:offset + 4])[0]
+                offset += 4
+                client_features = struct.unpack('!I', msg_data[offset:offset + 4])[0]
+                offset += 4
+            except (struct.error, IndexError):
+                pass
 
-            if not self.create_tunnel(address, uuid, remote_tunnel_id):
+            if not self.create_tunnel(address, uuid, remote_tunnel_id, client_features):
                 logger.warning("Failed to create tunnel (%s) while processing prepare request." % uuid)
                 self.write_message(address, CONTROL_TYPE_ERROR)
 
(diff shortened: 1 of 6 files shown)


RalfJung:
Oh, another thing I'd like to do: Have the client fail in `context_setup_tunnel` if the server returns a flag we do not know. That's clearly a non-conforming server which we do not want to have anything to do with.

mitar:
A side note: were there any changes on port usage? Can now multiple tunnels be on the same port? Maybe now that they fixed IDs, they can also support that? Or maybe we could send the patch upstream. Currently we have to use NAT ugliness to map different internal ports to same outside port, but there is nothing in the specs which would really require this. Kernel should differentiate between packets based on the ID, not based on the port.

If they fixed this, we could remove NAT and make everything much simpler and also support IPv6.

RalfJung:
I did not look into this at all.  The old NAT hacks certainly still work.

broker/l2tp_broker.cfg.example
```diff
@@ -21,6 +21,9 @@ namespace=default
 connection_rate_limit=10
 ; Set PMTU to a fixed value.  Use 0 for automatic PMTU discovery.
 pmtu=0
+; Whether this server runs a new kernel (4.13 and various stable series; in particular, 4.9.36) and
```
kostko: Could we perform some autodetection to determine this?

broker/src/tunneldigger_broker/protocol.py
```diff
@@ -112,37 +112,64 @@ def message(self, address, msg_type, msg_data, raw_length):
...
+            # 1 byte UUID length
+            # N bytes UUID
+            # optionally 4 bytes remote tunnel ID
+            # optionally 4 bytes client feature flags
+
+            offset = 0
+
             # Verify cookie value.
-            timestamp = msg_data[:2]
+            timestamp = msg_data[offset:offset+2]
```
kostko: Spaces around `+` operator.

broker/src/tunneldigger_broker/protocol.py
```diff
@@ -112,37 +112,64 @@ def message(self, address, msg_type, msg_data, raw_length):
...
+            offset += 2
             signed_value = '%s%s%s' % (address[0], address[1], timestamp)
             signature = hmac.HMAC(SECRET_KEY, signed_value, hashlib.sha1).digest()[:6]
             timestamp = struct.unpack('!H', timestamp)[0]
 
-            if signature != msg_data[2:8] or abs(protocol_time() - timestamp) > 2:
+            # Reject message if more than 2 protocol ticks old.  One tick is 1 >> 6 = 64 seconds.
+            if signature != msg_data[offset:offset+6] or abs(protocol_time() - timestamp) > 2:
```
kostko: Spaces around `+` operator.

broker/src/tunneldigger_broker/protocol.py
```diff
@@ -112,37 +112,64 @@ def message(self, address, msg_type, msg_data, raw_length):
...
             # Packet format:
             # 8 bytes of padding, to make sure the response is no larger than the request.
+            # optionally 4 bytes client feature flags
             if len(msg_data) < 8:
                 return
 
-            tunnel_manager = self.get_tunnel_manager()
+            client_features = 0
+            try:
+                client_features = struct.unpack('!I', msg_data[8:8+4])[0]
```
kostko: Spaces around `+` operator.

broker/src/tunneldigger_broker/tunnel.py
```diff
@@ -39,6 +39,10 @@
 PMTU_PROBE_REPEATS = 4
 PMTU_PROBE_COMBINATIONS = PMTU_PROBE_SIZE_COUNT * PMTU_PROBE_REPEATS
 
+# Session feature flags
+FEATURE_UNIQUE_SESSION_ID = 1 << 0
```
kostko: I think these should be moved into `protocol.py`.

broker/l2tp_broker.cfg.example
RalfJung: I guess we could -- but isn't that too fragile? I don't even have complete data which other kernels the "bad" patch was backported to.

RalfJung:
> try to figure out why we still have two clients connecting to the Gateway that now should report usage 0xFFFF.

I think I know what is happening. The problem is that these clients do not yet have <https://github.com/wlanslovenija/tunneldigger/commit/b367a33181fe6535e9866d855f9c2998aca2b531>. If they fail to get any reply from any broker, they will, by default, connect to the first one, which happens to be the one that already has the newer kernel.

broker/l2tp_broker.cfg.example
RalfJung: However, one thing we *could* do is detect the EEXIST on l2tpv3 session creation, and treat that as a hint that we need unique sessions IDs.  That would mean one client has a failed connection attempt, and old clients take 30s to re-try, but the next time around they would see usage 0xFFFF.

Would you prefer that?

RalfJung:
I have an updated version of this that automatically requires unique session IDs upon the first EEXIST on session creation. However, I also rebased on top of #67 (as I want them both on our servers).

I will update this PR once #67 got merged.

RalfJung:
We now have 10 nodes running with the new client, and 3 of our 4 gateways running with the new broker, for 4 days.  Everything seems to run smoothly. So as far as I am concerned, this is ready to be merged.

RalfJung:
Updated PR to contain the auto-detection, and rebased on top of master.

kostko:
Thanks!

PR #66, opened 2017-11-10
client: Quicker reconnect on broker failure

### LINKED ISSUE #65: Client behavior on broker restart is suboptimal
Currently, if the broker is restarted (i.e., sent a SIGINT and then started again), the following happens on the client:
* It starts out in STATE_KEEPALIVE, receives the error, switches to STATE_REINIT
* In context_process, this leads to calling context_reinitialize and then context_start_connect. In particular, `standby_only` gets set to true and the state switches to STATE_GET_COOKIE.
* The client sends a cookie request, and receives the response from the server that has come back up in the mean time.
* Upon receiving the reply, the client notices that `standby_only` is set, and does nothing.
* 30 seconds later, the main loop notices that the client has not been connected for 30 seconds, and restarts full broker selection, which eventually selects a sensible broker again.

This is probably not the intended behavior. However, I have a hard time figuring out the intended behavior, because there are two things in the main loop for detecting an error, and two things the other code does in case of error: The main loops checks for STATE_REINIT, and it starts a timer if the state is not STATE_KEEPALIVE. The other code sometimes goes to STATE_GET_COOKIE, sometimes to STATE_REINIT -- but in the latter case, context_process gets called before the main loop notices this state.

My inclination is to: (a) remove the STATE_REINIT check in the main loop; it doesn't work anyway; instead maybe the main connection timer should be lowered so that we get away from a bad broker faster. (b) make it so that when we receive an error in a running tunnel, we stick with this broker (like we do now), but do not set `standby_only` (which is clearly wrong). I think this requires another parameter to context_reinitialize saying whether the new context should be standby-only or not.

RalfJung:
This drops support for brokers that do not reply to a USAGE query. The advantage of that is that we can already tell from the state we are in whether we are in the "first phase" (scanning all brokers to see which ones are available) or the "second phase" (establishing and maintaining a tunnel with the broker).

Based on that, we can drop `standby_only`. We can also make sure that, once we are in the "second phase", if we ever jump to the "first phase" again, the main loop restarts the entire thing. I also decreased some timeouts to minimize the downtime in case a broker is shut down or restarted.

Before this patch, restarting a broker (if all other brokers are running) had a downtime of around 30s (until the main loop realizes that the connection is gone). This is despite the fact that the client got an ERROR, telling it that the broker went away! Stopping a broker (or restarting a broker while another broker is down) even lead to a downtime of 50s, because the client waits for 20s until it realizes that no reply is going to come to its USAGE query. With this patch, in both cases, the downtime is below 10s.

Fixes #65

### DIFF
--- a/client/l2tp_client.c
+++ b/client/l2tp_client.c
@@ -1386,6 +1386,10 @@ int main(int argc, char **argv)
     // Make sure all brokers are in sane state.
     for (i = 0; i < broker_cnt; i++) {
       context_reinitialize(brokers[i].ctx);
+      if (brokers[i].broken && brokers[i].broken + 3600 < timer_now()) {
+        // This one broke more than an hour ago, give it another chance.
+        brokers[i].broken = 0;
+      }
       if (!brokers[i].broken)
         working_brokers += 1;
     }
@@ -1455,7 +1459,7 @@ int main(int argc, char **argv)
     // Initially, we mark this broker as broken.  We will remove this mark after establishing
     // a connection.  We only want to consider a broker as broker if the initial connection fails;
     // disconnecting later (e.g. because the broker got restarted) is fine.
-    brokers[i].broken = 1;
+    brokers[i].broken = timer_now();
 
     // Perform processing on the main context; if the connection fails and does
     // not recover after 30 seconds, restart the broker selection process.
--- a/client/l2tp_client.c
+++ b/client/l2tp_client.c
@@ -122,7 +122,7 @@ enum l2tp_limit_type {
    When DNS resolving succeeds:
    -> STATE_GET_USAGE (sending a usage and a cookie request every 2s)
    when we receive usage information or a cookie
-   -> STATE_IDLE
+   -> STATE_STANBDY
 
    Now broker selection is performed; for the selected broker the main loop changes the state so
    that we go on:
@@ -144,7 +144,7 @@ enum l2tp_ctrl_state {
   STATE_REINIT,
   STATE_RESOLVING,
   STATE_GET_USAGE,
-  STATE_IDLE,
+  STATE_STANBDY,
   STATE_GET_COOKIE,
   STATE_GET_TUNNEL,
   STATE_KEEPALIVE,
@@ -208,9 +208,6 @@ typedef struct {
   // Tunnel uptime.
   time_t tunnel_up_since;
 
-  // Whether we received usage information from the broker
-  int standby_available;
-
   // PMTU probing.
   int pmtu;
   int peer_pmtu;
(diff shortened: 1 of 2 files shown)


RalfJung:
The CI fails, it seems it actually tests that the client still works with old brokers.

Considering that typically, the brokers are much easier to update than the client (AFAIK), is it acceptable to drop this compatibility? Old clients should still work with new brokers, as this only affects the client's state machine.

RalfJung:
I reworked the patches to maintain support with old servers, and rebased on top of current master.

RalfJung:
@kostko Should I rebase again? I can keep doing that, but so far I got no feedback whatsoever on this PR.

mitar:
I am not sure about the content, but please do rebase because this will also test against v0.3.0 which you added and was just merged, no?

RalfJung:
Good point. Done.

RalfJung:
Looking good :)

client/l2tp_client.c
```diff
@@ -1386,6 +1419,10 @@ int main(int argc, char **argv)
     // Make sure all brokers are in sane state.
     for (i = 0; i < broker_cnt; i++) {
       context_reinitialize(brokers[i].ctx);
+      if (brokers[i].broken && brokers[i].broken + 3600 < timer_now()) {
```
kostko: Should the retry period be made configurable instead of hardcoded to 3600 seconds?

client/l2tp_client.c
```diff
@@ -1449,46 +1484,43 @@ int main(int argc, char **argv)
...
         syslog(LOG_ERR, "Connection to %s lost.", main_context->broker_hostname);
         break;
-      } else if (main_context->state == STATE_KEEPALIVE) {
-        // We successfully established a connection, this broker is fine.
-        brokers[i].broken = 0;
       }
 
       // If the connection is lost, we start the reconnection timer.
-      if (restart_timer && main_context->state != STATE_KEEPALIVE) {
+      // Hitting this code path should not be possible, but lets play safe.
```
kostko: Can you add an explanation why in the comment?

client/l2tp_client.c
```diff
@@ -115,15 +115,40 @@ enum l2tp_limit_type {
   LIMIT_TYPE_BANDWIDTH_DOWN = 0x01
 };
 
+/* The state machine looks as follows:
```
kostko: :+1:

client/l2tp_client.c
RalfJung: This is not the only timeout -- there's also the 20s (I think) for the initial availability check, and the 10s for restarting the entire thing.

I considered making them all macros, but did not want the PR to get bigger than necessary. Making them CLI arguments is of course also an option, but is that really useful?

client/l2tp_client.c
RalfJung: Done.

RalfJung:
Any chance of getting this through this week? I'd prefer not to do the firmware update over Christmas. :D

PR #68, opened 2017-11-20
 Add autotools for building the client

kaechele:
## Commit message
This enables easier building for different targets, e.g. LEDE/OpenWrt or
other Linux distributions.

Also enables building with shared libasyncns, which is turned on by
default if the shared version is found on the system.
For this the libasyncns source files were moved to their own
subdirectory so that the include statement will pull in the header file
for the shared version instead of the bundled one.

I also made the decision to check in all required generated files and
not include them in the gitignore file. This should make releasing and
building releases much easier. However, this means that changes in
either configure.ac or Makefile.am require an autoreconf and subsequent
check-in of the newly generated files.

## Rationale
In Freifunk Networks it can be desirable to have full fledged Linux clients (sometimes acting as servers) connect to the network. To more easily facilitate this it is helpful to use the original tunneldigger client.
Using autotools, while indeed adding complexity and some cruft to the repo, we can build the client out of the same sources used for the LEDE/OpenWrt package more elegantly.
Autotools automatically determines whether libnl or libnl-tiny is used and whether libasyncns is available as a shared library in the build system. This lays the foundation to having officially packaged version of the client in distribution repos (such as Fedora's) that require unbundling of libraries while maintaining the option to statically include libasyncns for size reasons when building for embedded platforms.

I indent submitting a RPM spec file to the repo for easy packaging in a later pull request.

Tested on:
- Fedora 27 (x86_64)
- LEDE master (bc50a97dfc9f6bf3c1c9d0abc56d8ae56b5d48a1) on TP-Link Archer C25v1

Looking forward to your comments.


kostko:
I think it would be better to use CMake instead of autotools?

kaechele:
I'll resubmit something based on cmake. I believe I have found my error.

kaechele:
Obsoleted by #69

PR #76, opened 2018-01-03
docs: Add instructions for building client

robimarko:
Add instructions for building Tunneldigger client under Ubuntu/Debian.
This will really help a lot of newcomers

### DIFF
--- a/docs/client.rst
+++ b/docs/client.rst
@@ -13,6 +13,28 @@ This will give you a ``tunneldigger`` directory which contains the broker and
 the client in separate directories. Client code can be compiled into a
 stand-alone program.
 
+In order to compile client code following packages are required on Ubuntu/Debian::
+
+    build-essential pkg-config cmake libnl-genl-3-dev libnl-3-dev libasyncns-dev
+    
+Root permission are needed for package installation and Tunneldigger client installation.
+
+Installation
+------------
+
+If we assume that you are currently in ``/tunneldigger/client`` folder you can do::
+
+    cmake -G "Unix Makefiles"
+    
+After configuration has been succesfully built you can simply::
+
+    make
+    
+After code has been compiled in order to install simply::
+
+    sudo make install
+    
+
 OpenWrt Package
 ---------------
 


RalfJung:
When would it ever be useful to `sudo make install`? That will just pollute the user's system... I think usually, running the binary directly from the build dir makes more sense.

robimarko:
Well,most users and me really dont like having to know the path to binary or go to the directory where it is just to run it.

I can make a note that that is not needed and that it can be run in the folder in which it is compiled

kaechele:
Is the `cmake -G "Unix Makefiles"` really required? Usually a `cmake .` should suffice, if you're trying to do a quick build.

robimarko:
Well,I assume you can skip it,but its recommended so why not just copy/past

valentt:
Any chance to merge these instructions, because they are very helpful.

docs/client.rst
```diff
@@ -13,6 +13,28 @@ This will give you a ``tunneldigger`` directory which contains the broker and
...
+    build-essential pkg-config cmake libnl-genl-3-dev libnl-3-dev libasyncns-dev
+    
+Root permission are needed for package installation and Tunneldigger client installation.
+
+Installation
+------------
+
+If we assume that you are currently in ``/tunneldigger/client`` folder you can do::
+
+    cmake -G "Unix Makefiles"
```
RalfJung: the `-G ...` is unnecessary AFAIK, I have never seen it in build instructions.

Also, cmake prefers out-of-source builds. Why do you build in the source dir here?

docs/client.rst
```diff
@@ -13,6 +13,28 @@ This will give you a ``tunneldigger`` directory which contains the broker and
...
+    
+After configuration has been succesfully built you can simply::
+
+    make
+    
+After code has been compiled in order to install simply::
+
+    sudo make install
```
RalfJung: I think this last part should be removed. `sudo make install` is probably not that useful. It is certainly not needed for development/testing.

PR #80, opened 2018-06-25
allow fractions of seconds for the rate limit

RalfJung:
I have this now set to 0.1 seconds on our servers, and that works very well to distribute floods of clients (occurring when another server goes down).


robimarko:
@RalfJung Wlanslovenija is pretty much dead so I doubt that anybody will merge this.

RalfJung:
Hm, they were still fine ~6 months ago...

robimarko:
After that whole Wlanslovenija is down to one man which is really busy.

PR #86, opened 2019-01-08
pass the local broker port to the session setup scripts

RalfJung:
We are using the local broker port (i.e., the one that the clients connect to) to distinguish which of several networks the client is going to be added to.  To do that, we need this information to be available in the session-up script.

### DIFF
--- a/broker/scripts/setup_interface.sh
+++ b/broker/scripts/setup_interface.sh
@@ -1,7 +1,13 @@
 #!/bin/bash
 TUNNEL_ID="$1"
+SESSION_ID="$2"
 INTERFACE="$3"
 MTU="$4"
+ENDPOINT_IP="$5"
+ENDPOINT_PORT="$6"
+LOCAL_PORT="$7"
+UUID="$8"
+LOCAL_BROKER_PORT="$9"
 
 . scripts/bridge_functions.sh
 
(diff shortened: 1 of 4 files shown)


RalfJung:
Cycling PR to re-run CI.

kaechele:
As I and others are working on getting rid of the NAT voodoo we are currently using to workaround the Kernel limitations with regards to L2TP Port / Session ID usage: Is this the right way to approach this? As soon as we get rid of the NAT stuff there will be no local broker port anymore.

RalfJung:
Well, I have a real need for this kind of information right now.  We cannot do multi-domain in our site without this patch, so we already deployed it.

If you think getting rid of NAT will land within weeks, I'm happy to close this as the situation will resolve itself soon.  If you think it is going to be several months, it might be worth landing this now instead of making everyone wait (or carry local patches).

kaechele:
I think I misunderstood what you meant here, sorry. The local port of the the broker (which is the one that the client uses to connect and is the same for all clients using the broker) is obviously not going away 😅

RalfJung:
Nobody complained for long enough, I am going to merge this. :)

RalfJung:
*sigh* Travis/GH network is so flaky today :/

PR #88, opened 2019-01-08
fix logic for unbreaking brokers

RalfJung:
Namely, make sure that if there are no candidates left (whether they are offline or marked broken), we definitely re-enable all brokers.

Fixes #87

PR #91, opened 2019-04-07
Update l2tp_client.c

valentt:


RalfJung:
This patch has already landed with https://github.com/wlanslovenija/tunneldigger/pull/88.

PR #93, opened 2019-06-17
Document how to find out the linux kernel version

niccokunzmann:
This lowers the threshold for beginners

### DIFF
--- a/docs/server.rst
+++ b/docs/server.rst
@@ -36,7 +36,7 @@ Prerequisites
 -------------
 
 The first thing you need is a recent (>= 2.6.36) Linux kernel that supports L2TPv3
-tunnels. You can find out your linux kernel version using the command ``uname -r``.
+tunnels. You can find out your linux kernel version using the command ``uname -a``.
 
 The following modules are required for Tunneldigger operation:
 
--- a/docs/server.rst
+++ b/docs/server.rst
@@ -36,7 +36,9 @@ Prerequisites
 -------------
 
 The first thing you need is a recent (>= 2.6.36) Linux kernel that supports L2TPv3
-tunnels. The following modules are required for Tunneldigger operation:
+tunnels. You can find out your linux kernel version using the command ``uname -r``.
+
+The following modules are required for Tunneldigger operation:
 
 * ``l2tp_core``
 * ``l2tp_eth``


RalfJung:
Isn't `uname -v` a better suggestion? On Debian `-r` shows the Debian package version, `-v` the upstream version.

RalfJung:
Also cycling PR to make CI re-run, looks like a spurious failure.

kaechele:
Fedora:
```
[felix@x1 ~]$ uname -r
5.1.8-300.fc30.x86_64
[felix@x1 ~]$ uname -v
#1 SMP Sun Jun 9 17:09:32 UTC 2019

```
Ubuntu:
```
felix@ubuntu:~$ uname -r
5.0.0-13-generic
felix@ubuntu:~$ uname -v
#14-Ubuntu SMP Mon Apr 15 14:59:14 UTC 2019
```

RalfJung:
The Ubuntu thing doesn't exactly look like an upstream kernel version either. `5.0.0-13` is not a thing.

Debian:
```
$ uname -r
4.19.0-5-amd64
$ uname -v
#1 SMP Debian 4.19.37-3 (2019-05-15)
```
"4.19.37" is the actual upstream version.

Maybe just recommend `uname -a`?

niccokunzmann:
I do not know if this is the case for you with the failing PRs: I had a configuration where it failed for external people without write access to the repository because the secrets if Travis did not resolve.

I pushed uname -a because that works for me, too and seems more universal

RalfJung:
Oh wow, Travis is getting better every day. At producing ridiculous situations. Now even `git checkout $TRAVIS_COMMIT` is failing.

niccokunzmann:
It does a checkout. https://travis-ci.org/wlanslovenija/tunneldigger/jobs/548248390#L436
Maybe it would be sufficient to print what the head commit is and what the pull request commit is. 
I also wondered: Fetching the commits from the "origin" may not allow fetching commits from my repository. So, I would not expect it to run.

Would it be ok if we just echo the current commit?

RalfJung:
> Fetching the commits from the "origin" may not allow fetching commits from my repository.

GH integrates PR commit into the PR the PR is submitted to. You can even fetch them by querying the right ref: `git fetch origin pull/PRNUMBER/head`.

RalfJung:
But I guess we'll just have to ignore `TRAVIS_COMMIT`, and instead query the top commit.

PR #94, opened 2019-06-17
document loading modules

niccokunzmann:
this makes it easier for me and beginners

### DIFF
--- a/docs/server.rst
+++ b/docs/server.rst
@@ -84,10 +84,6 @@ If the modules were loaded successfully, your listing of the modules might look 
     ip6_udp_tunnel         16384  1 l2tp_core
     udp_tunnel             16384  1 l2tp_core
 
-In addition the kernel must support network address translation via netfilter,
-otherwise the tunnels will not work as Tunneldigger uses translation to achieve
-that all tunnels operate over the same external port.
-
 Also, if you want to have working bandwidth limits, the kernel must support traffic
 shaping and the HTB queuing discipline, together with ``fq_codel``.
 
--- a/docs/server.rst
+++ b/docs/server.rst
@@ -38,6 +38,15 @@ Prerequisites
 The first thing you need is a recent (>= 2.6.36) Linux kernel that supports L2TPv3
 tunnels. You can find out your linux kernel version using the command ``uname -a``.
 
+These distributions are supported:
+
+* Debian
+* Fedora with the package ``kernel-modules-extra``
+* *add your distribution*
+
+Kernel Modules
+~~~~~~~~~~~~~~
+
 The following modules are required for Tunneldigger operation:
 
 * ``l2tp_core``
--- a/docs/server.rst
+++ b/docs/server.rst
@@ -69,7 +69,7 @@ Manual Activation of the Kernel Modules
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 You can also activate the modules using ``modprobe``.
-A restart will load the kernel without them if you do not create the file.
+A system restart will *not* load the modules again if you do not create the file as described above.
 
 .. code:: shell
 
--- a/docs/server.rst
+++ b/docs/server.rst
@@ -39,7 +39,7 @@ The first thing you need is a recent (>= 2.6.36) Linux kernel that supports L2TP
 tunnels. You can find out your linux kernel version using the command ``uname -a``.
 
 We assume the following instructions to work on the distributions listed below.
-You are welcome to add your distribution if the instructions work and to edit them.
+You are welcome to add your distribution if the instructions work and to edit them to make them work.
 
 * Debian
 * Fedora with the package ``kernel-modules-extra``
--- a/docs/server.rst
+++ b/docs/server.rst
@@ -95,15 +95,6 @@ your listing of the modules might look like this:
     ip6_udp_tunnel         16384  1 l2tp_core
     udp_tunnel             16384  1 l2tp_core
 
-Additional Notes About Kernel Modules
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-Also, if you want to have working bandwidth limits, the kernel must support traffic
-shaping and the HTB queuing discipline, together with ``fq_codel``.
-
-The system should be configured to load these modules at boot which is usually done
-by listing the modules in ``/etc/modules``.
-
 System Packages
 ---------------
 
--- a/docs/server.rst
+++ b/docs/server.rst
@@ -44,6 +44,26 @@ The following modules are required for Tunneldigger operation:
 * ``l2tp_eth``
 * ``l2tp_netlink``
 
+You can find out if these modules are loaded by running ``lsmod | grep l2tp``.
+If you get no output, they need to be activated.
+
+.. code:: shell
+
+    $ sudo modprobe l2tp_core
+    $ sudo modprobe l2tp_eth
+    $ sudo modprobe l2tp_netlink
+
+Then, your listing of the modules can look like this:
+
+.. code:: shell
+
+    $ lsmod | grep l2tp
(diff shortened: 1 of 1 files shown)


RalfJung:
Beginners are extremely unlikely to encounter a system where this is even needed. See https://github.com/wlanslovenija/tunneldigger/pull/96#issuecomment-502816967: I think we should recommend creating a file in `/etc/modules-load.d` instead.

Doing `modprobe` will just mean that people will wonder why it stopped working after a reboot.

kaechele:
How detailed would we need the documentation to be?
Generally, if this level of detail is required, would it maybe make more sense to say which Distributions are supported?
So for example, Fedora needs an extra package for the modules: kernel-modules-extra

niccokunzmann:
> How detailed would we need the documentation to be?

I would go with the documentation that is there and add what I would have needed.

docs/server.rst
```diff
@@ -36,12 +36,54 @@ Prerequisites
...
+The following modules are required for Tunneldigger operation:
 
 * ``l2tp_core``
 * ``l2tp_eth``
 * ``l2tp_netlink``
 
+You can find out if these modules are loaded by running ``lsmod | grep l2tp``.
+If you get no output, they need to be activated.
+For the activation of the kernel modules, we recommend adding a file
+``/etc/modules-load.d/tunneldigger.conf`` with the following content:
```
RalfJung: People should just always add that. Just because `lsmod` lists these now, doesn't mean they will still be there after a reboot or after some unrelated configuration changed.

docs/server.rst
```diff
@@ -36,12 +36,54 @@ Prerequisites
...
+You can also activate the modules using modprobe.
+A restart will load the kernel without them if you do not create the file.
+
+.. code:: shell
+
+    $ sudo modprobe l2tp_core
+    $ sudo modprobe l2tp_eth
+    $ sudo modprobe l2tp_netlink
+
+Then, your listing of the modules can look like this:
```
RalfJung:
```suggestion
If the modules were loaded successfully, your listing of the modules can look like this:
```

docs/server.rst
```diff
@@ -36,12 +36,54 @@ Prerequisites
...
+    $ lsmod | grep l2tp
+    l2tp_eth               16384  0 
+    l2tp_netlink           24576  1 l2tp_eth
+    l2tp_core              32768  2 l2tp_eth,l2tp_netlink
+    ip6_udp_tunnel         16384  1 l2tp_core
+    udp_tunnel             16384  1 l2tp_core
+
 In addition the kernel must support network address translation via netfilter,
 otherwise the tunnels will not work as Tunneldigger uses translation to achieve
 that all tunnels operate over the same external port.
```
RalfJung: There is no distro wirth mentioning that does not support netfilter, right?

I think I'd like to see this and the next 2 paragraphs removed. They are just confusing people by pointing out a problem that is extremely unlikely to exist.

docs/server.rst
niccokunzmann: Wow, how do you add this suggestion?

docs/server.rst
niccokunzmann: 1d208e5 covers this

docs/server.rst
niccokunzmann: 4cddd02 restructures this and puts the activation first

docs/server.rst
RalfJung: Magic. ;)

When adding in-line comments in a PR, there's a button above the text box that looks like a document icon with "+ -" in it. That lets you write in-line suggestions that you can accept directly in the web UI.

docs/server.rst
```diff
@@ -32,26 +32,80 @@ the broker.
...
+
+    l2tp_core
+    l2tp_eth
+    l2tp_netlink
+
+Manual Activation of the Kernel Modules
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+You can also activate the modules using ``modprobe``.
+A restart will load the kernel without them if you do not create the file.
```
RalfJung:
```suggestion
A system restart will *not* load the modules again if you do not create the file as described above.
```

docs/server.rst
```diff
@@ -32,26 +32,80 @@ the broker.
...
+    udp_tunnel             16384  1 l2tp_core
+
+Additional Notes About Kernel Modules
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Also, if you want to have working bandwidth limits, the kernel must support traffic
 shaping and the HTB queuing discipline, together with ``fq_codel``.
 
 The system should be configured to load these modules at boot which is usually done
 by listing the modules in ``/etc/modules``.
```
RalfJung: This should be changed to recommend adding them to the file above. Or maybe just add them right there in the docs?

docs/server.rst
```diff
@@ -32,26 +32,80 @@ the broker.
...
-Prerequisites
--------------
+Operating System
+----------------
 
 The first thing you need is a recent (>= 2.6.36) Linux kernel that supports L2TPv3
-tunnels. The following modules are required for Tunneldigger operation:
+tunnels.
+
+These distributions are supported:
```
RalfJung: It's not that tunneldigger doesn't work on other distributions, it's just that we don't have instructions. Maybe reword this to say that? Something like "The following instructions should work for these distributions:"

niccokunzmann:
Thanks!

docs/server.rst
niccokunzmann: (1) Until now, I do not understand this:

> Also, if you want to have working bandwidth limits, the kernel must support traffic
shaping and the HTB queuing discipline, together with ``fq_codel``.

(2) And to me, this

> The system should be configured to load these modules at boot which is usually done
by listing the modules in ``/etc/modules``.

looks like a duplication of the *Kernel Module Activation on Boot* Section and can be removed.

docs/server.rst
niccokunzmann: 26bd3b3 addresses this.

docs/server.rst
kaechele: fq_codel is the default qdisc for any newer Distribution running systemd. So it's mostly built into the kernel on those distributions. If not it will appear as a kernel module called sch_fq_codel.
Again, I see no need to explicitly state how to create backwards compatibility here since every major distribution has this by default. If you are running a very old and unsupported release of a distribution like Fedora 21, which is now over 4 years old, you probably have a good reason for that and an understanding of how to backport changes to make stuff work. Just mentioning fq_codel in this case will give you enough hints to successfully google a solution to your problem. The tunneldigger documentation shouldn't be the place to teach people how to perform basic Linux administration tasks.
Other than that I do really appreciate the clarity that is being added to the documentation on the other parts. Nice job @niccokunzmann!

niccokunzmann: @kaechele Thanks! I wonder: We could make a section which states what old (4years+) or self-built kernels need in order to run Tunneldigger. If I can read that as a precondition for the section, I can skip this section in normal mode. Would anyone of you appreciate movong the other deleted content to that section?

RalfJung: I think removing it is just fine.

Do you want to remove the mention of `/etc/modules` here, or should I merge and then do it?

RalfJung:
Hm, I cannot merge due to merge conflicts. Can you resolve those?

niccokunzmann:
This is my view of it:
image
I see no merge conflicts. Do you mean that I should merge master first before you can merge it because you have a policy to only accept fast-forward merges?

RalfJung:
I get

> This branch cannot be rebased due to conflicts
> Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch. 

So please rebase on top of current master.

niccokunzmann:
I did a rebase and a force push.

Issue #95, opened 2019-06-17
Documentation about netfilter


niccokunzmann:
The [documentation](https://tunneldigger.readthedocs.io/en/latest/server.html) states:

> In addition the kernel must support network address translation via netfilter, otherwise the tunnels will not work as Tunneldigger uses translation to achieve that all tunnels operate over the same external port.

[Here](https://www.netfilter.org/), I read that this is included in Kernel 2.4.x+ to have this if we have iptables installed.

> netfilter.org is home to the software of the packet filtering framework inside the Linux 2.4.x and later kernel series. Software commonly associated with netfilter.org is iptables.

As I am new to the project in such internals, I am a bit unsure: If I have iptables on my computer, will it work? Will it work if I just have a kernel later than 2.4.x?

kaechele:
Can be closed

RalfJung:
Which PR fixed this?

kaechele:
As per your comment here: https://github.com/wlanslovenija/tunneldigger/pull/96#issuecomment-504133692
you stated that this should be addressed in #94. The author of #96 agreed and closed #96. #94 is merged.
Interestingly #96 actually removed any mention of netfilter other than in the package names.

All this leads me to infer this can be closed.

Also, this will be obsoleted (hopefully soon) by #115

PR #96, opened 2019-06-17
Describe check for netfilter

niccokunzmann:
fixes https://github.com/wlanslovenija/tunneldigger/issues/95

### DIFF
--- a/docs/server.rst
+++ b/docs/server.rst
@@ -45,6 +45,8 @@ tunnels. The following modules are required for Tunneldigger operation:
 In addition the kernel must support network address translation via netfilter,
 otherwise the tunnels will not work as Tunneldigger uses translation to achieve
 that all tunnels operate over the same external port.
+If you have a Linux kernel of version 2.4.x or later, this is included.
+You can still check with the command ``sudo iptables --version``.
 
 Also, if you want to have working bandwidth limits, the kernel must support traffic
 shaping and the HTB queuing discipline, together with ``fq_codel``.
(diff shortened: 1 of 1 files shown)


kaechele:
This information is irrelevant.
L2TP was introduced in 2.6.23. The only reason you might not have netfilter enabled in any Kernel that would support tunneldigger is because you built it yourself and have it purposely disabled. If you did that you know all about it.

RalfJung:
TBH it might be best to just remove most of that discussion of kernel modules, and assume a default configuration. AFAIK all distros have everything enabled we need. So even talking about this is just confusing.

RalfJung:
Well okay we still need the part about loading them.

So maybe just tell people to create an `/etc/modules-load.d/tunneldigger.conf` containing
```
l2tp_core
l2tp_eth
l2tp_netlink
```

niccokunzmann:
I see these lines in the .travis.yml:

https://github.com/wlanslovenija/tunneldigger/blob/8d4d2f9abeb2644e943e7f4babb885960aecb5d8/.travis.yml#L17
https://github.com/wlanslovenija/tunneldigger/blob/8d4d2f9abeb2644e943e7f4babb885960aecb5d8/.travis.yml#L18
I think they are enabled by default.

However, it might not hurt adding them next to the others of they are needed.

RalfJung:
This is overlapping with https://github.com/wlanslovenija/tunneldigger/pull/94. I suggest to close this PR and do the netfilter edits there.

niccokunzmann:
:+1:

PR #98, opened 2019-06-18
Add heading for packages

niccokunzmann:
The prerequisites have two sections:
kernel modules and packages
This heading clarifies this together with
https://github.com/wlanslovenija/tunneldigger/pull/94/files

### DIFF
--- a/docs/server.rst
+++ b/docs/server.rst
@@ -52,6 +52,9 @@ shaping and the HTB queuing discipline, together with ``fq_codel``.
 The system should be configured to load these modules at boot which is usually done
 by listing the modules in ``/etc/modules``.
 
+System Packages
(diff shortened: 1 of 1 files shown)


RalfJung:
Could you also add a similar head for the "Kernel modules" part? Seems more consistent to have them both in sub-sectons then.

niccokunzmann:
@RalfJung Yes, I will. #94 does that. If I do it in an additional PR, I get merge conflicts. I wanted to avoid that.

RalfJung:
Then I suggest we just close this PR and you do the other heading change in #94, too. That also give me a chance to see how the file will look like when you are done with it. ;)

niccokunzmann:
done. 
4cddd02

PR #105, opened 2019-07-02
clarify ebtables' module character

niccokunzmann:
I noticed this as part of the docker image: ebtables must be activated.


RalfJung:
Thanks!

Issue #92, opened 2019-05-03
Python2: End of life is coming soon - port tunneldigger-broker to python3


pmelange:
Python2 will see it's end-of-life coming at the end of the year.  https://pythonclock.org/

Most specifically, OpenWrt has a discussion going on about dropping python2 at https://github.com/openwrt/packages/issues/8520

It would be great if tunneldigger-broker were to be ported/updated to python3.

pmelange:
There is now an issue in OpenWrt to transition all python2 programs to python3 https://github.com/openwrt/packages/issues/8893

kaechele:
Can be closed

RalfJung:
Indeed, fixed by https://github.com/wlanslovenija/tunneldigger/pull/110 :)

PR #110, opened 2019-09-30
Python 3 Support

kaechele:
This is a first stab at Python 3 support.
From my testing, it should be complete.
Coverage report from running the broker with the default config and connecting a client as well as connecting a client with a set bandwidth limit:

```
Name                 Stmts   Miss  Cover
----------------------------------------
__init__.py              0      0   100%
_ffi/__init__.py         0      0   100%
broker.py              132     29    78%
conntrack.py            76     13    83%
eventloop.py            26      1    96%
genetlink.py            59      1    98%
hooks.py                91      8    91%
l2tp.py                106     29    73%
limits.py               28      9    68%
main.py                 63     17    73%
netlink.py             160     24    85%
network.py              88     18    80%
protocol.py             95     12    87%
timerfd.py              80     25    69%
traffic_control.py      15      1    93%
tunnel.py              184     53    71%
----------------------------------------
TOTAL                 1203    240    80%
```
Areas that are uncovered are mainly untriggered error states.

Tested and working on:
```
python2-2.7.16-4.fc31.x86_64
python3-3.7.4-5.fc31.x86_64
```

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -1,6 +1,13 @@
 vNEXT
 -----
 
+* Python 3 support.
+  Python 2 is no longer supported.
+  **NOTE:** If you are running the broker in a Python virtualenv you will have to
+  rebuild the virtualenv using a Python 3 interpreter.
+  This can be achieved by deleting the old virtualenv folder and recreating it.
+  For recreating the virtualenv you can refer to the "Installation" section of
+  the broker `installation chapter in the Tunneldigger documentation`_.
 * Fixed compatibility with new Linux kernels on the broker side: New kernels
   force the l2tpv3 session ID to be unique system-wide, while old tunneldigger
   clients have a hard-coded ID of 1 for both ends of the tunnel. When a new
@@ -16,6 +23,8 @@ vNEXT
   probes.
 * Improve client behavior on broker failure, and decrease some reconnect
   timeouts.
+
+.. _`installation chapter in the Tunneldigger documentation`: https://tunneldigger.readthedocs.io/en/latest/server.html#installation
 
 v0.3.0, 2017-Apr-02
 -------------------
--- a/broker/setup.py
+++ b/broker/setup.py
@@ -29,7 +29,6 @@ setup(
     ],
     install_requires=[
         'netfilter>=0.6.2',
-        'six>=1.10.0',
         'cffi>=1.4.1',
     ],
     extras_require={},
--- a/broker/setup.py
+++ b/broker/setup.py
@@ -21,6 +21,7 @@ setup(
         'Intended Audience :: System Administrators',
         'License :: OSI Approved :: GNU Affero General Public License v3',
         'Programming Language :: Python',
+        'Programming Language :: Python :: 3 :: Only',
     ],
     include_package_data=True,
     zip_safe=False,
--- a/broker/src/tunneldigger_broker/broker.py
+++ b/broker/src/tunneldigger_broker/broker.py
@@ -3,7 +3,7 @@ import socket
 import time
 import traceback
 
-import conntrack
+from . import conntrack
 import netfilter.table
 import netfilter.rule
 
@@ -42,7 +42,7 @@ class TunnelManager(object):
         self.hook_manager = hook_manager
         self.max_tunnels = max_tunnels
         self.tunnel_id_base = tunnel_id_base
-        self.tunnel_ids = set(xrange(tunnel_id_base, tunnel_id_base + max_tunnels))
+        self.tunnel_ids = set(range(tunnel_id_base, tunnel_id_base + max_tunnels))
         self.tunnel_port_base = tunnel_port_base
         self.namespace = namespace
         self.tunnels = {}
@@ -117,11 +117,11 @@ class TunnelManager(object):
             self.last_tunnel_created = now
         except KeyboardInterrupt:
             raise
-        except l2tp.L2TPTunnelExists, e:
+        except l2tp.L2TPTunnelExists as e:
             # Do not return the tunnel identifier into the pool.
             logger.warning("Tunnel identifier %d already exists." % e.tunnel_id)
             return False
-        except l2tp.L2TPSessionExists, e:
+        except l2tp.L2TPSessionExists as e:
             # Return tunnel identifier into the pool.
             self.tunnel_ids.add(tunnel_id)
             logger.warning("Session identifier %d already exists." % e.session_id)
@@ -208,7 +208,7 @@ class TunnelManager(object):
         manager instance should not be used after calling this method.
         """
 
-        for tunnel in self.tunnels.values():
+        for tunnel in list(self.tunnels.values()):
             try:
                 tunnel.close()
             except:
--- a/docs/server.rst
+++ b/docs/server.rst
@@ -137,10 +137,18 @@ If we assume that you are installing Tunneldigger under ``/srv/tunneldigger``
 (the scripts provided with Tunneldigger assume that as well), you can do::
 
     cd /srv/tunneldigger
-    virtualenv env_tunneldigger
+    virtualenv -p /usr/bin/python3 env_tunneldigger
 
-This creates the virtual environment. You can then checkout the Tunneldigger
-repository into ``/srv/tunneldigger/tunneldigger`` by doing::
+.. note::
+    Tunneldigger only supports Python 3.
+    
+    Using the above command ensures the virtualenv is created using a Python 3 interpreter.
+    In case the Python 3 interpreter you would like to use is not located at ``/usr/bin/python3`` you will have to adjust the path accordingly.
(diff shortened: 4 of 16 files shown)


kaechele:
Addresses #92

kaechele:
I am now running this version on one of our gateways: https://map.fflev.de/#!/en/map/02cafe760101
So far no errors in the log and the connections are working fine.

broker/src/tunneldigger_broker/broker.py
```diff
@@ -208,7 +208,7 @@ def close(self):
         manager instance should not be used after calling this method.
         """
 
-        for tunnel in self.tunnels.values():
+        for tunnel in list(self.tunnels.values()):
```
mweinelt: the list cast looks unnecessary, because you are already calling `.values()`, which gets you `dict_values([])`, which is iterable as well.

kaechele: If we were dropping Python 2 completely, then yes, we could leave it out.
See: https://python-future.org/compatible_idioms.html#dict-keys-values-items-as-a-list
I guess this could be another step when refactoring this whole thing.

kaechele: After thinking about it a bit more: I think you are right. Since we're directly iterating over it and don't actually need it to be a list instead of a dictview I can drop the list cast.

kaechele: I actually put this back in.
If you operate directly on the values you will eventually end up with `RuntimeError: dictionary changed size during iteration` when stopping the broker. So operating on a copy with list() is helpful.

RalfJung:
The error log says
```
Downloading/unpacking netfilter==0.6.2 (from -r /srv/tunneldigger/broker/requirements.txt (line 1))

  Cannot fetch index base URL https://pypi.python.org/simple/

  Could not find any downloads that satisfy the requirement netfilter==0.6.2 (from -r /srv/tunneldigger/broker/requirements.txt (line 1))

Cleaning up...

No distributions at all found for netfilter==0.6.2 (from -r /srv/tunneldigger/broker/requirements.txt (line 1))

Storing debug log for failure in /home/travis/.pip/pip.log
```
Do you know what that means? I've not used pip besides installing something by hand interactively...

kaechele:
> That sounds like something that might be caused by this PR?

No, it's not. If you trigger a build for master it will fail with the same errors. Tried it myself here: https://travis-ci.org/kaechele/tunneldigger/jobs/597459907

Looks like our testing setup is broken. It is wildly outdated too.

RalfJung:
Well it worked until a few months ago. So something must have changed somewhere... and because Python doesn't have proper lockfiles, old builds are not reproducible. :(

RalfJung:
@kaechele could you rebase, then we can see what CI says now that it is based on Ubuntu Bionic.

kaechele:
CI failed as expected since it uses Python 2 exclusively. The point of this change is to drop Python 2 support.
So for CI we would now either replace Python 2 with Python 3 or (and I very much discourage that) add another layer to run all the tests for Python 2 and Python 3 basically creating a large amount of test cases run by CI.
So a decision is needed here.

RalfJung:
Note that our CI also tests current clients against old servers and vice versa. So the CI harness anyway will have to be able to still run python 2 brokers. That's why [this file](https://github.com/wlanslovenija/tunneldigger/blob/f5bb0efb19468e8dfd0726b40c1431f5c526b3b7/tests/run_server.sh) knows how to run various versions of the broker.

So, basically, I think that file (and probably also [the setup script](https://github.com/wlanslovenija/tunneldigger/blob/f5bb0efb19468e8dfd0726b40c1431f5c526b3b7/tests/prepare_server.sh)) should detect Python-3-based brokers and then use Python 3 to launch them instead of Python 2. I wouldn't run the *same* broker with both Python 2 and Python 3 though.

kaechele:
Success :)
It probably wouldn't be too much work to make this fully Python 2 compatible as well. But I don't see a reason to do that.

RalfJung:
Nice! Did you do this conversion by hand or with a tool, i.e., how likely is it that some rarely taken branch of tunneldigger still does something that won't work any more? Given Python's dynamic nature, I am a bit worried about the side-effects of this.

> It probably wouldn't be too much work to make this fully Python 2 compatible as well. But I don't see a reason to do that.

Well I guess the question is, do we want to officially support both Python versions for some time (maybe to make migration easier)?

Also, I assume the docs at `docs/server.rst` will need updating?

broker/setup.py
```diff
@@ -29,7 +30,6 @@
     ],
     install_requires=[
         'netfilter>=0.6.2',
-        'six>=1.10.0',
```
RalfJung: Why can this be removed?

kaechele: This is a Python3 to Python2 compatibility layer. Unless we want to support Python 2 as well we don't need six.

kaechele:
> Nice! Did you do this conversion by hand or with a tool, i.e., how likely is it that some rarely taken branch of tunneldigger still does something that won't work any more? Given Python's dynamic nature, I am a bit worried about the side-effects of this.

The initial conversion was done by the 2to3 tool: https://github.com/wlanslovenija/tunneldigger/pull/110/commits/246ebc54412f4b5ba4e5d4c7b5598ddd339bfdf7
Then I did some manual edits from hand after testing: https://github.com/wlanslovenija/tunneldigger/pull/110/commits/70bcf6ff9edfa188801d7353f97ba573f87f9345

Especially the UTF-8 encoding changes from Py2 to Py3 required some manual testing. But as you can see in the initial comment on this PR checked the code coverage of the tests I perfomed and the were nearly complete.

I expect very little side effects.

> > It probably wouldn't be too much work to make this fully Python 2 compatible as well. But I don't see a reason to do that.
> 
> Well I guess the question is, do we want to officially support both Python versions for some time (maybe to make migration easier)?

How would that make migration easier? All distributions that run Python 2 also run Python 3. It's a matter of rebuilding your virtualenv, if you use that. If you installed it into the system path it's even just a simple reinstall.
So I don't understand which migration path you would be referring to.

> Also, I assume the docs at `docs/server.rst` will need updating?
Yes, I guess. I'll look into that next.

RalfJung:
> So I don't understand which migration path you would be referring to.

I guess it's about the virtualenv then. So the migration path is simply to `rm -rf` the previous virtualenv, and then rebuild it as a Python 3 virtualenv?

kaechele:
> So the migration path is simply to `rm -rf` the previous virtualenv, and then rebuild it as a Python 3 virtualenv? That should go into `HISTORY.rst`.

Exactly. Given that virtualenv and Python 3 are already available (which is the case on nearly any Linux Distribution that still receives updates).

virtualenv -p enables you to select which Python interpreter should be used (as can be seen in the CI related commit in this set). From within the virtualenv the versionless python interpreter in the $PATH will then refer to the version of the interpreter that was used to create the venv.

I agree that we should do a good cleanup of both the docs and `HISTORY.rst`. There are a few changes that didn't make it in there.

RalfJung:
> I agree that we should do a good cleanup of both the docs and HISTORY.rst. There are a few changes that didn't make it in there.

A proper cleanup can wait, but for this PR I'd like to see at least a brief note with the instructions for how to upgrade in `HISTORY.rst`, and the [`virtualenv` call in the docs](https://github.com/wlanslovenija/tunneldigger/blob/f5bb0efb19468e8dfd0726b40c1431f5c526b3b7/docs/server.rst#L140) be updated, so that virtualenv noobs like me can manage the upgrade.

HISTORY.rst
```diff
@@ -1,6 +1,13 @@
 vNEXT
 -----
 
+* Python 3 support.
+  Python 2 is no longer supported.
+  **NOTE:** If you are running the broker in a Python virtualenv you will have to
+  rebuild the virtualenv using a Python 3 interpreter.
+  This can be achieved by deleting the old virtualenv folder and recreating it.
+  For recreating the virtualenv you can refer to the "Installation" section of
+  the broker installation chapter in the Tunneldigger documentation.
```
RalfJung: Can you turn this into a link?

docs/server.rst
```diff
@@ -139,8 +139,16 @@ If we assume that you are installing Tunneldigger under ``/srv/tunneldigger``
     cd /srv/tunneldigger
     virtualenv env_tunneldigger
```
RalfJung: Might make sense to add a `-p` here as well?

RalfJung:
@kaechele  This is basically ready to land, except for the two nits I noted.
(No rush, I just saw you being active on other PRs so I was not sure we agreed on the status of this PR.)

docs/server.rst
kaechele: Reworded this whole thing in the new push.

HISTORY.rst
kaechele: Done

PR #111, opened 2019-09-30
client: declare project as C

kaechele:
This prevents cmake from looking for a C++ compiler.

### DIFF
--- a/client/CMakeLists.txt
+++ b/client/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 2.8)
-project (tunneldigger)
+project (tunneldigger C)
 
 set(USE_LIBNL FALSE CACHE BOOL "Explicitly use libnl over libnl-tiny")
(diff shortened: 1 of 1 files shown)


RalfJung:
Hm, no, actually it's some Python thing...
```
No distributions at all found for netfilter==0.6.2 (from -r /srv/tunneldigger/broker/requirements.txt (line 1))
```
I'll have to investigate, and don't have time for that right now (and maybe this came up in some of the other PRs).

RalfJung:
@kaechele I think after a rebase, CI should be green now.

kaechele:
Seems like the CI is still not bulletproof.
But I thank you for your persistence to fix this pesky CI stuff :D

RalfJung:
Yeah, Travis has really bad network connectivity so downloading the LXC image frequently fails even with the 10 retries we are now doing...

mitar:
BTW, maybe consider using  GitHub Actions? They are available for this repo as well.

PR #99, opened 2019-06-18
[WIP] Server Docker Image

niccokunzmann:
This is work in progress. I would like to create a version of the server image which can be set up with a low skill level and only a few required options.

If you have suggestions, they are welcome. If I fail, this is a pull request which people can look at later.

TODO:

- [x] how to configure olsrd
- [ ] start the image
- [ ] connect to it with a valid installation
- [ ] create automated build or delete the section about providing the image [comment](https://github.com/wlanslovenija/tunneldigger/pull/99/files/75095f220bf03ea07845d24861d0dd9da635e121#r295940349)
- [X] clarify if OLSR is needed/nice to have/useless [comment](https://github.com/wlanslovenija/tunneldigger/pull/99#discussion_r295939518)
- [ ] add meaning of [comment](https://github.com/wlanslovenija/tunneldigger/pull/99#discussion_r295985587) in the [routing deamon](https://tunneldigger.readthedocs.io/en/latest/server.html#routing-daemon) Section
- [x] remove OLSR
- [ ] add DHCP for client and broker

PR #112, opened 2019-10-19
bump Ubuntu version we use on CI

RalfJung:
I'll use this PR to debug our CI problems.

@kaechele you mentioned already trying this an running into LXC issues. Do you have a branch for that? Also you suggested to use Vagrant instead for testing; that's fine for me but my Vagrant experience is about on the same level as my LXC experience (which is zero). We also have a [WIP docker image](https://github.com/wlanslovenija/tunneldigger/pull/99) that might help, maybe.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,15 +19,16 @@ install:
   - sudo modprobe nf_conntrack_netlink
   - sudo iptables -t nat -L >/dev/null
   - sudo iptables -t filter -P FORWARD ACCEPT
-  - sudo /usr/bin/python3 ./tests/tunneldigger.py --setup
+  - sudo /usr/bin/python3 ./tests/tunneldigger.py --setup bionic
+  - sudo /usr/bin/python3 ./tests/tunneldigger.py --setup trusty
 script:
   - sudo -E sh ./tests/travis.sh
 
 env:
   matrix:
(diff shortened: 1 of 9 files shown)


kaechele:
I've done some testing here: https://travis-ci.org/kaechele/tunneldigger

The problem is with the newer Ubuntu versions that the cloning of the LXC container fails and the Python library for LXC has exactly zero error handling or logging. That's about how far I got.

kaechele:
Okay. So I got as far as getting the LXC stuff somewhat working again: https://travis-ci.org/kaechele/tunneldigger/builds/601569731

But now we're facing issues with the older versions of the client not compiling due to errors thrown by the newer compiler.
Also there is no coverage for any Python 3 testing.

RalfJung:
> But now we're facing issues with the older versions of the client not compiling due to errors thrown by the newer compiler.

That is odd; I am running this tunneldigger broker on several Debian Stretch servers just fine... which is newer than xenial.

>  Also there is no coverage for any Python 3 testing.

My rough plan was to first get CI work again with the old broker (Python 2 only), then make some kind of release or snapshot or so, and then move to Python 3 and drop Python 2 support. Supporting both at the same time does not seem worth the hassle, does it?

RalfJung:
Yay, we have working CI again. :)

PR #114, opened 2019-12-28
Client: prefix log messages with broker host+port

RalfJung:


RalfJung:
@mitar any chance you could have a look at this? Or @kaechele do you feel confident reviewing client changes?

lynxis:
@RalfJung can you describe you 'slight cleanup' more detailed? What are you doing in here? In general I would do 2 PR.
For the logging: can you create a #define for the logging. E.g.
```
LOGXYZ(broker, level, fmt, args...) syslog(level, "[%s:%s] " fmt, broker->host, borker->port, ## args)
```

RalfJung:
I don't think I agree with the advice for the macro; C macros are very fragile and error-prone and I'd rather avoid them. Given that we'd need two versions of the macro (one for brokers and one for contexts), I don't think they provide sufficient simplification of the codebase.

> can you describe you 'slight cleanup' more detailed? What are you doing in here? In general I would do 2 PR.

Fair, though I feel just reviewing the two commits separately also works.

What I am doing here is a small refactoring to remove the useless (and needlessly complicated) `broker_select_one` function. In C, an array of length 1 and a single element are the same thing, so we can just use `broker_select` with length 1. I also added some comments, because "broker_select" initially sounded to me like it was about selecting which broker to use, and not about the `select` syscall.

kaechele:
I'd like to suggest using more descriptive commit messages than "slight cleanup".

lynxis:
The "slight cleanup" commit is fine, except for the commit message.

> C macros are very fragile and error-prone and I'd rather avoid them

I agree, macros are fragile and macros shouldn't be used much, but in this case i would use them. It's a small macros, it's easy to understand. Using a macro here instead of a function is much simpler. A function would be more overhead.

I would use 2 macros here, one for context, one for broker.

RalfJung:
I adjusted the first commit's message.

> It's a small macros, it's easy to understand. Using a macro here instead of a function is much simpler. A function would be more overhead.

These are variadic macros we are talking about... not sure if they qualify as "small", but certainly they do not qualify as simple. [This](https://en.wikipedia.org/wiki/Variadic_macro) is quite the horror show I got to say.^^

I am not sure when I will have the time to figure out how these work, to do the experiment if that really makes the code cleaner. But if you want to try this, I'd be happy to incorporate your commit into this PR (you can even make a PR against this PR branch if you want).

RalfJung:
@mitar are you okay landing this as-is or would you prefer me to figure out how variadic macros work?

mitar:
I do not care, I am not really looking into the code anymore here. Just trying to help with some high-level ideas/feedback.

PR #116, opened 2020-01-02
Initial IPv6 support

### LINKED ISSUE #75: Get it working with IPv6
It would be really great to be able to use native IPv6 on the WAN side, since more and more broadband-consumers get "IPv6 + DSlite".
Meaning that IPv4 is only "via CarrierGradeNat/CGN", where we have to observe very often strange effects like dying connections and/or packet loss. While IPv6 is working fine. 

As far as i understand, it's not a simple job like for IPv4, since the packet header sizes are not that static in V6, but it would be really helpful, even it would end up as a total rewrite.

kaechele:
This implements IPv6 support for both broker and client.
The broker implementation is very straightforward.
The client is a little more involved.

Initial testing showed mixed results. My initial testing suggests there may be an MTU issue with a particular ISP in our region when using IPv6. So further testing would be appreciated.

Depends on #115 and #110 
Fixes #75


kaechele:
I will rework and resubmit this. This doesn't cleanly rebase or merge onto master by now anyway.

RalfJung:
@kaechele great! Indeed I think we can consider the other patches this is based on to be "battle-tested" by now and hence can move forward.

I wonder if we should have a release of some sort...

PR #118, opened 2020-01-20
Update top-level docs

kaechele:
Apart from changing the links in the license files to HTTPS the README.rst
got updated:
- improved structure
- removal of obsolete links to wlan-si
- mention GitHub as primary development platform
- include license info
- include contributions info
- remove listing of individual contributors
- refer to GitHub insights for authorship info

**NOTE:** The removal of the individual contributors may be controversial. However, since that list will probably always be outdated anyway, I recommend just referring people to the commit stats of the repository.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,8 +1,13 @@
 Tunneldigger
 ============
 
+L2TPv3 VPN tunneling solution
+
 .. image:: https://travis-ci.org/wlanslovenija/tunneldigger.svg?branch=master
     :target: https://travis-ci.org/wlanslovenija/tunneldigger
+
+About
+-----
 
 Tunneldigger is one of the projects of `wlan slovenija`_ open wireless network.
 It is a simple VPN tunneling solution based on L2TPv3 tunnels supported in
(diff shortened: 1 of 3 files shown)


mitar:
I like this. I agree about removal of the contributors.

README.rst
```diff
@@ -1,52 +1,65 @@
...
+-----
+
 Tunneldigger is one of the projects of `wlan slovenija`_ open wireless network.
 It is a simple VPN tunneling solution based on L2TPv3 tunnels supported in
 recent Linux kernels.
 
 .. _wlan slovenija: https://wlan-si.net
 
-Documentation is found at:
+It consists of a client and server portion and is optimized to run on embedded
```
RalfJung: Only the client is really optimized for that, I'd say.

This might also be a good opportunity to introduce the "broker" terminology, like

> consists of a client and a server/"broker" portion

README.rst
```diff
@@ -1,52 +1,65 @@
...
+We welcome code and documentation contributions to Tunneldigger in the form of
+`Pull Requests`_ on GitHub where they can be reviewed and discussed by the
+community.
+We encourage everyone to check out any pending pull requests and offer comments
+or ideas as well.
+
+.. _Pull Requests: https://github.com/wlanslovenija/tunneldigger/pulls
+
+Tunneldigger is developed by a community of developers from many different
+backgrounds and has seen significant contributions from the community.
```
RalfJung: The repeated use of "community" makes this sound a bit odd. I think this sounds better if you remove the second half (after "and"), but if you can find a better formulation that'd also be fine.

README.rst
kaechele: Fixed.

README.rst
kaechele: I just dropped the second part. Seemed kinda redundant when I read it again today.

README.rst
```diff
@@ -1,52 +1,66 @@
...
+-----
+
 Tunneldigger is one of the projects of `wlan slovenija`_ open wireless network.
 It is a simple VPN tunneling solution based on L2TPv3 tunnels supported in
 recent Linux kernels.
 
 .. _wlan slovenija: https://wlan-si.net
 
-Documentation is found at:
+It consists of a client and server portion referred to as the broker. The
```
RalfJung:
```suggestion
It consists of a client and a server portion referred to as the broker. The
```

kaechele: Fixed

mitar:
Just one late comment, the description currently states:

"L2TPv3 VPN tunneling solution"

Maybe this should be expanded, to explain that it is using UDP and standard Linux kernel stuff. So it is performant and stable? Or something. But UDP at least is missing there.

Also GitHub project description is currently:

"Client and broker for our custom L2TPv3 NAT-traversing tunnel setup protocol based on L2TPv3 support in the Linux kernel."

I think that should be also improved (dropping "our" and "custom" maybe)? Also it could be synced up with the description from README.

PR #119, opened 2020-01-31
explain at least some bits in the magic prefix

RalfJung:
Source: @kaechele at https://github.com/wlanslovenija/tunneldigger/pull/115#issuecomment-573668028


kaechele:
LGTM

PR #121, opened 2020-02-04
update and sync top-level docs

RalfJung:
@kaechele's PR only updated README.rst, I did not realize that docs/index.rst is basically a duplication of the same information. This syncs them, with some further minor changes.

Also Cc @mitar


kaechele:
I like this. Reads much more concise this way.

RalfJung:
All right, merging then. :)

Issue #106, opened 2019-07-17
client: suspicious assertion in asyncns.c


RalfJung:
Building the client local shows a warning that seems legit at first glance:
```
In file included from /home/r/src/freifunk/tunneldigger/client/libasyncns/asyncns.c:23:
/home/r/src/freifunk/tunneldigger/client/libasyncns/asyncns.c: In function ‘asyncns_setuserdata’:
/home/r/src/freifunk/tunneldigger/client/libasyncns/asyncns.c:1484:12: warning: suggest parentheses around assignment used as truth value [-Wparentheses]
     assert(q->asyncns = asyncns);
            ^
/home/r/src/freifunk/tunneldigger/client/libasyncns/asyncns.c: In function ‘asyncns_getuserdata’:
/home/r/src/freifunk/tunneldigger/client/libasyncns/asyncns.c:1492:12: warning: suggest parentheses around assignment used as truth value [-Wparentheses]
     assert(q->asyncns = asyncns);
            ^
```
This needs further investigation to determine if the `=` should be `==`.

kaechele:
I contacted Lennart Poettering on October 27, 2019 but haven't heard back yet. This quite clearly looks like a typo.

lynxis:
I've sent the patch to Lennart via mail.

RalfJung:
Would be really good to hear back from Lennart about this -- or anyone else familiar with the asyncns code, actually.

Issue #70, opened 2017-11-22
tunneldigger dies somtimes


charlemagnelasse:
Original bug report from @Sunz3r under https://github.com/freifunk-gluon/gluon/issues/1188

---

If internet lost connection (like reboot of router or plug off lan-cable) then the tunneldigger sometime dies.After strace the tunneldigger i found that a assert is triggered: https://github.com/wlanslovenija/tunneldigger/blob/master/client/asyncns.c#L884https://pastebin.com/R6CQmGXr

---

I was not able to report this bug on there site ...

this bug happens when dns-resolution failed and despite sending data through a socket:

     td-client: Failed to send() control packet (errno=89, type=1)!

when dns-resolution suddenly works in this situation then the tunneldigger will crash.

I think there is a bug in the broker selection. The selector should wait for successful address-resolution.

To workaround this bug you can use ip's instead of names.

mitar:
Have you verified this is still the issue with latest tunneldigger? Because many things around broker selection were improved.

But yes, I think this code path with DNS resolution is less tested because we are not using it ourselves, we have hard-coded IPs.

charlemagnelasse:
@Sunz3r: Please provide some feedback

RalfJung:
We are using DNS names and I have not seen this yet. But the issue is not fully reproducible anyway if I understand the report correctly?

Btw, there is an error in the links in the first post, it seems there are two links directly stuck to each other.

RalfJung:
> I think there is a bug in the broker selection. The selector should wait for successful address-resolution.

I don't follow. Of course it does that, the selector cannot contact a server before address resolution completed.

Isn't this a bug in libasyncns? This is an internal assertion of that library: https://github.com/wlanslovenija/tunneldigger/blob/master/client/libasyncns/asyncns.c#L884

PR #122, opened 2020-02-05
client: fix warning suspicious assertion in asyncns

lynxis:
It seems to be a typo in the check as all other function check for equality between
both arguments.

Fixes: #106


RalfJung:
It *seems* to be a typo, I agree. But I'd prefer to be more sure than that, and the code *seems* to be working fine right now.

kaechele:
> and the code _seems_ to be working fine right now.

Does it though?
https://github.com/wlanslovenija/tunneldigger/issues/70
:D

RalfJung:
The thing is, if the assertion is true, the assignment doesn't change anything.

PR #123, opened 2020-02-05
l2tp_client: allow to set MAX_BROKERS from outside

scherke:
For certain environments 10 brokers are not enought. Allow to
overwrite this value.
make CFLAGS="-D MAX_BROKERS=254"

### DIFF
--- a/docs/client.rst
+++ b/docs/client.rst
@@ -23,3 +23,10 @@ package. Source code for such OpenWrt package can be `found here`_.
 .. _OpenWrt: https://openwrt.org/
 
 You can add the whole repository as an OpenWrt feed and add package to your firmware.
+
+Configuration
+-------------
+
+* **MAX_BROKERS** (default: 10): Maximum number of brokers that can be handled in a single process.
+
+    make CFLAGS="-D MAX_BROKERS=20"
--- a/client/l2tp_client.c
+++ b/client/l2tp_client.c
@@ -236,7 +236,9 @@ typedef struct {
 } broker_cfg;
 
 // Maximum number of brokers that can be handled in a single process.
-#define MAX_BROKERS 10
+#ifndef MAX_BROKERS
+  #define MAX_BROKERS 10
+#endif
 
 // Forward declarations.
 void context_delete_tunnel(l2tp_context *ctx);


RalfJung:
Could you add this in some suitable place in the documentation?

docs/client.rst
```diff
@@ -23,3 +23,15 @@ package. Source code for such OpenWrt package can be `found here`_.
...
 You can add the whole repository as an OpenWrt feed and add package to your firmware.
+
+Configuration
+-------------
+
+* **MAX_BROKERS** ``default: 10`` Maximum number of brokers that can be handled in a single process.
+
+    make CFLAGS="-D MAX_BROKERS=20"
+
+or you can edit the ``Makeflie`` and add
```
RalfJung: There's no `Makefile` (at least none that anyone should edit by hand), the client is using cmake.

Also, I am not sure what the right way is to format this with reStructuredText, but the "or you can edit" part is not appropriately indented like the part before (you can [see it rendered here](https://github.com/wlanslovenija/tunneldigger/blob/2ef04245c913a741826eea4e6715265f2131c83e/docs/client.rst)).

scherke: I mean the packages-opkg [``Makefile``](https://github.com/wlanslovenija/firmware-packages-opkg/tree/master/net/tunneldigger) 
I deleted the "or ..."

docs/client.rst
```diff
@@ -23,3 +23,10 @@ package. Source code for such OpenWrt package can be `found here`_.
 .. _OpenWrt: https://openwrt.org/
 
 You can add the whole repository as an OpenWrt feed and add package to your firmware.
+
+Configuration
+-------------
+
+* **MAX_BROKERS** ``default: 10`` Maximum number of brokers that can be handled in a single process.
```
RalfJung:
```suggestion
* **MAX_BROKERS** (default: 10) Maximum number of brokers that can be handled in a single process.
```
I don't see a good reason to use code formatting here.

scherke: Okay, then that's how it's done. Do I have to change this in my pull request?

RalfJung: You should be able to press a button in GH's UI to accept my suggestion.

scherke: I don't have the button **Commit suggestion**. I will change my commit manually.

RalfJung: That's odd... it always worked for me. Thanks for doing it manually!

Issue #126, opened 2020-03-06
Some Kernels have broken SO_REUSEPORT handling, only one Tunnel connection possible


kaechele:
> This bug is meant for future reference.

Newer versions of Tunneldigger use `SO_REUSEPORT` to process multiple tunnels on one single port. In the past Tunneldigger used a NAT-based workaround to make this work. To simplify the code and remove unnecessary dependencies this workaround was removed.
Unfortunately there was a bug in several versions of the Kernel that exhibited a behaviour where packets would show up in random sockets rather than the one dedicated to the specific 4-tuple.
This means that the change in conjunction with the bug has some peculiar implications for which Kernel versions can be used.

**Generally, if you run any currently supported Linux distribution with all updates installed you shouldn't run into this issue.**
You have probably landed here because you still use an older Linux distribution or haven't updated to a working Kernel version.
If you are experiencing this issue you have two options:

1. Update your Kernel to a supported version or upgrade your Distribution to one that has a supported Kernel version.
2. Revert to an older Tunneldigger version that uses the NAT workaround.

The following Kernel versions will **not** work:

- For the 3.x series: 3.9 and older (no `SO_REUSEPORT` support)
- For the 4.x series: 4.4 to 4.19.75
- For the 5.x series: 5.2.17 and older

Any other Kernel should work but we didn't test them individually.
We have tested Debian 10, Fedora 31 and Ubuntu 19.10 and all of those work.

RalfJung:
Debian oldstable currently only offers kernels that are affected by this bug; I [submitted a bugreport](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=953291) about updating the backport kernels (it's already a 4.19 kernel, but not new enough).

PolynomialDivision:
Could u say something what happens with old kernels? I updated the version to 19.07.4 with 4.14.180 or 4.19.X kernel.
I don't see any obvious issues with that kernel and the version while running tunneldigger. I will investigate further.

RalfJung:
Clients were unable to reliably connect with broken kernels, so we saw many connection timeouts or other disconnects in the logs. Also see https://github.com/wlanslovenija/tunneldigger/issues/129. You should also see warnings specifically pointing out that the kernel is likely buggy.

Maybe Ubuntu backported the problematic patches, who knows.

PolynomialDivision:
[...]

PR #127, opened 2020-03-06
Reintroduce: Remove NAT logic (v2)

kaechele:
Re-submission of https://github.com/wlanslovenija/tunneldigger/pull/115; see [here](https://github.com/wlanslovenija/tunneldigger/pull/115#issuecomment-580992574) for some relevant discussion.

Added note and link to issue description regarding the `SO_REUSEPORT` bug in some Kernel versions.
Otherwise just reverted the revert, meaning that the NAT removal code from the initial pull request has been re-introduced unmodified. That should make a re-review redundant.


broker/src/tunneldigger_broker/tunnel.py
```diff
@@ -350,7 +288,7 @@ def create_tunnel(self, address, uuid, remote_tunnel_id, client_features):
         """
 
         if address != self.endpoint:
-            logger.warning("Protocol error: tunnel endpoint has changed.")
+            logger.warning("Protocol error: tunnel endpoint has changedi. Possibly due to kernel bug. See: https://github.com/wlanslovenija/tunneldigger/issues/126")
```
RalfJung:
```suggestion
            logger.warning("Protocol error: tunnel endpoint has changed. Possibly due to kernel bug. See: https://github.com/wlanslovenija/tunneldigger/issues/126")
```

docs/server.rst
```diff
@@ -35,7 +35,7 @@ the broker.
 Operating System
 ----------------
 
-The first thing you need is a recent (>= 2.6.36) Linux kernel that supports L2TPv3
+The first thing you need is a recent (>= 3.9) Linux kernel that supports L2TPv3
```
RalfJung: In [the issue](https://github.com/wlanslovenija/tunneldigger/issues/126) you say "3.9 and older" does not work, but this one says "3.9 and newer" will work. So does 3.9 itself work or not? The two statements are contradiction each other in that regard.

RalfJung: The issue also claims that 4.19.75 and 5.2.17 don't work even though you said elsewhere those are the first fixed version, so this is probably just an issue of you using `<= version` instead of `< version` in the issue text.

kaechele: Totally agree.
I updated the issue to reflect your comments. Please check if it makes more sense that way.

broker/src/tunneldigger_broker/tunnel.py
kaechele: You can clearly tell I use vim :yum: 
Fixed and force-pushed.

docs/server.rst
RalfJung: I did some more edits that IMO clarify things further, let me know what you think.
Given that the leading version number in Linux (3, 4, 5) is utterly meaningless, I'd prefer to group things by major kernel versions (x.y).

kaechele: Yes. That is much better now. Thanks!

RalfJung: Actually from what I could find, the fix is in 5.4 but not in 5.3, so I adjusted the issue accordingly.

RalfJung: Ah, it got backported to 5.3.1.

RalfJung:
I adjusted the issue some more: from what I could see, the [faulty commit](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=e32ea7e74727) landed in 4.5, but the 4.4 kernel should be unaffected.

RalfJung:
I'm afraid this still doesn't work here... see https://github.com/wlanslovenija/tunneldigger/issues/129.

PR #128, opened 2020-03-07
client: dont send PMTU_NTFY when we do not actually have a sensible PMTU

RalfJung:
I noticed clients are sending CONTROL_TYPE_PMTU_NTFY with silly PMTUs (0 - L2TP_TUN_OVERHEAD overflows). This PR should stop that.

### DIFF
--- a/client/l2tp_client.c
+++ b/client/l2tp_client.c
@@ -1199,13 +1199,13 @@ void context_process(l2tp_context *ctx)
           ctx->pmtu = ctx->probed_pmtu;
           context_session_set_mtu(ctx);
         }
-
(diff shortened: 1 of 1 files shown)


RalfJung:
@kaechele any chance you could test this as well? Then I would submit an update to Gluon, to start getting #114 and this downstream.

RalfJung:
@kaechele any chance you could have a look at this?

kaechele:
Tested this today. LGTM.
Please rebase and squash out the merge commit and we have a deal ;-)

RalfJung:
I *think* that happens automatically when I hit "rebase and merge"... but maybe I should not risk that. ;)

PR #130, opened 2020-03-07
broker: without automatic PMTU, tell client about static PMTU

RalfJung:
this informs the client about our MTU so both sides use the same


RalfJung:
@kaechele what do you think about this? How do you handle MTUs in your network?

mitar:
Why would automatic PMTU not work?

mitar:
So Tunneldigger has a custom implementation.

RalfJung:
We talked about this a while ago in https://github.com/wlanslovenija/tunneldigger/issues/62. Note that tunneldigger already has support for disabling auto-PMTU, this just improves that support.

I also think that for our use-case, automatic PMTU is rather useless: we are running a batman mesh network on top of tunneldigger, and the entire mesh is a single layer 2 domain. This means the entire mesh needs to have a single MTU. I do not see how determining precisely the MTU of each tunnel is useful when on top of that tunnel, we anyway need to work with a single MTU that works for everyone. That's why we statically configure a suitably small tunneldigger MTU in our network, which seems to work.

mitar:
I see. Thanks.

RalfJung:
All right, merging then. We've been using this in production for a few weeks now.

PR #131, opened 2020-03-07
check for endpoint change on all messages, not just CREATE_TUNNEL

RalfJung:
This check can be done for all messages, and with broken kernels it can indeed go wrong for all messages, to check it properly.


kaechele:
Please squash the two merge commits. Otherwise LGTM.

Issue #129, opened 2020-03-07
Incoming messages get incorrectly dispatched to broker


RalfJung:
We re-landed @kaechele's NAT removal, but things are still going very wrong: once I have more than just 1 or 2 clients (I am not sure if there is a fixed limit, but with 2 clients things still seemed to work fine), connections start to drop. The server thinks that the client timed out.

I added some extra logging, and from what I can see, incoming messages from the client often arrive at the (2-tuple) broker socket, not at the (4-tuple) tunnel socket. So despite what @kaechele said [here](https://github.com/wlanslovenija/tunneldigger/pull/115#issuecomment-581429934), there does not seem to be a guarantee that messages with a matching 4-tuple socket do indeed get delivered to that socket.

This is on Debian buster:
```
$ uname -a
Linux gw1.saar.freifunk.net 4.19.0-8-amd64 #1 SMP Debian 4.19.98-1 (2020-01-26) x86_64 GNU/Linux
```

RalfJung:
When using the backports kernel, the issues goes away.
```
$ uname -a
Linux gw1.saar.freifunk.net 5.4.0-0.bpo.3-amd64 #1 SMP Debian 5.4.13-1~bpo10+1 (2020-02-07) x86_64 GNU/Linux
```
However, requiring a 5.4 kernel seems a bit excessive...

RalfJung:
I also tried some [awful hacks](https://github.com/freifunk-saar/tunneldigger/tree/nonat) that only use one socket per broker. However, then I get "OSError: Netlink error: Device or resource busy (16)" when the second tunnel is created. Maybe the kernel does not support more than one l2tp tunnel per socket?

Also the tunnel didn't actually seem to transport any data, but that could easily be some silly mistake somewhere in my hacks.

RalfJung:
I'll make a guess and say that [this commit](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=4cdeeee9252af1ba50482f91d615f326365306bd) could be the one fixing the bug here. It first appeared in 5.0, and unfortunately does not seem to have been backported.

kaechele:
> I'll make a guess and say that [this commit](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=4cdeeee9252af1ba50482f91d615f326365306bd) could be the one fixing the bug here. It first appeared in 5.0, and unfortunately does not seem to have been backported.

[...]

PR #135, opened 2020-03-17
Better error and docs for our kernel requirements

RalfJung:
Fixes https://github.com/wlanslovenija/tunneldigger/issues/129

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -24,8 +24,11 @@ vNEXT
 * Improve client behavior on broker failure, and decrease some reconnect
   timeouts.
 * Removed dependency on Netfilter.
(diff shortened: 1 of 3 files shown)


kaechele:
I'll deploy this today. I'm a little concerned with performance as we loop over every tunnel for each message we receive.

RalfJung:
> I'll deploy this for testing today.

Thanks!

> I'm a little concerned with performance as we loop over every tunnel for each message we receive.

We only do that for messages we receive on the broker, i.e., for new clients. Messages that go to a tunnel socket don't get any checking.

kaechele:
This has been running on one of our gateways in production now for the last week. No issues.
LGTM

RalfJung:
Awesome, thanks for testing!

Issue #136, opened 2020-03-21
[Docs] max_tunnels


yanosz:
Heiho,

for configuration, users choose an upper limit for `max_tunnels`
https://github.com/wlanslovenija/tunneldigger/blob/master/broker/l2tp_broker.cfg.example#L9

From a UX perspective, presenting a configuration option requires the user to choose a valid value.
Nevertheless, neither the comment in the sample config nor https://tunneldigger.readthedocs.org/ help to do so

IMHO the docs should explain, in what way platform (CPU cores, memory, etc.) and kernel perameters (i.e. sysctl.conf) influce reasonable configurations.

If 1024 is a sane configuration in almost all situations - I've no experience, here - it will be helpful to document this.

RalfJung:
> IMHO the docs should explain, in what way platform (CPU cores, memory, etc.) and kernel perameters (i.e. sysctl.conf) influce reasonable configurations.

Yeah, that would be great. Unfortunately, I have no idea about this.

Empirically, our VMs struggle really hard with 256 tunnels, so that's the limit we are using. But this depends not just on server details, it also depends on the typical client workload. I doubt there is general advice we can give here, TBH.

yanosz:
Good point - I've two ideas:

* If 256 is good in your situation, it'd cool to document that. (Hardware, network design, etc.)
* I've no idea to what extent the client workload actually influences the system load - but, that doesn't really matter, imho. If you go down the rabbit hole, you probably end-up using OVS instead of a simple linux bridge to implement QoS (htb, fair queues or simple rate limits). That's out of scope - imho. Let's just document configurations, that are known to be working.

yanosz:
sorry, wrong button

PR #137, opened 2020-03-22
tweak and document default settings

RalfJung:
I am also fine leaving the default tunnel limit at 1024, but then we should likely give an example for the kind of machine where that makes sense.

For the `connection_rate_limit`, I think 10 is a really bad default value. It means that even with just 100 clients, after a broker restart it takes 1000 seconds for all clients to reconnect! We are using 0.1 in production here.

Fixes https://github.com/wlanslovenija/tunneldigger/issues/136
Cc @yanosz

### DIFF
--- a/broker/l2tp_broker.cfg.example
+++ b/broker/l2tp_broker.cfg.example
@@ -5,13 +5,15 @@ address=127.0.0.1
 port=53,123,8942
 ; Interface with that IP address
 interface=lo
-; Maximum number of tunnels that will be allowed by the broker
-max_tunnels=1024
+; Maximum number of tunnels that will be allowed by the broker.
+; On a cheap VPS, more than 256 tunnels usually do not make sense.
+; Physical machines can take more.
+max_tunnels=256
 ; Tunnel id base
 tunnel_id_base=100
 ; Reject connections if there are less than N seconds since the last connection.
 ; Can be less than a second (e.g., 0.1).
-connection_rate_limit=10
+connection_rate_limit=0.2
 ; Set PMTU to a fixed value.  Use 0 for automatic PMTU discovery.  A non-0 value also disables
 ; PMTU discovery on the client side, by having the server not respond to client-side PMTU
 ; discovery probes.


RalfJung:
> ooks ok. I'd cool if you could be more specific on the VM and bare metal host
(e.g. Amazon EC2 T3 nano - year 2020, or whatever host you use).

The VMs we are using are on various European hosters, nothing standardized. I doubt they can be compared to other hosters or even within the same hoster, these offers change rapidly and they don't tell you the details of the hardware or how many customers there are per physical machine. On the plus side, they are really cheap. ;)

I don't think there is any useful way to name "reference VMs" here.

yanosz:
Thanks for your comment and effort. Maybe other people can share their experience, too. IMHO: Having a small doc section on hardware requirements (i.e. size of running installations) appears to be realistic.

RalfJung:
@kaechele @mitar what do you think about these changes to the default settings?

broker/l2tp_broker.cfg.example
```diff
@@ -5,13 +5,15 @@ address=127.0.0.1
...
+; Maximum number of tunnels that will be allowed by the broker.
+; On a cheap VPS, more than 256 tunnels usually do not make sense.
+; Physical machines can take more.
+max_tunnels=256
 ; Tunnel id base
 tunnel_id_base=100
 ; Reject connections if there are less than N seconds since the last connection.
 ; Can be less than a second (e.g., 0.1).
-connection_rate_limit=10
+connection_rate_limit=0.2
```
mitar, line 16: Hm, maybe we should have rate limiting per source IP, not global. Otherwise rate limiting is really hard.

mitar:
Looks good to me.

broker/l2tp_broker.cfg.example
RalfJung: @mitar  You mean something like https://github.com/wlanslovenija/tunneldigger/pull/138? :D

mitar: :-) So let's update then defaults then. :-)

PR #138, opened 2020-03-27
add connection rate limit per IP address (with custom hook)

RobWei:
Hello,

I have implemented a feature that triggers a hook for exceeding a connection rate per IP address.
The value in connection_rate_limit_per_ip sets the time in seconds a client is not allowed to reconnect without triggering the hook.

The need for the implentation were clients that caused endless connection requests, one after another and caused maximum cpu consumption. On the other hand it could also be a targeted Denail of Service.

Bildschirmfoto vom 2020-03-27 14-23-16
Bildschirmfoto vom 2020-03-27 14-22-11

We will use the hook to block an IP address over iptables for 5 minutes. I have chosen the hook to allow for more flexibility in script design.


RalfJung:
Thanks for the PR! I have seen high load due to repeated connection attempts as well in Freifunk Saar, so some rate limiting here might indeed make sense.

> wants to merge 2 commits into wlanslovenija:newkernel

Please make your PRs against the master branch, not against feature branches.

broker/src/tunneldigger_broker/broker.py
```diff
@@ -82,6 +85,15 @@ def create_tunnel(self, broker, address, uuid, remote_tunnel_id, client_features
         if self.last_tunnel_created is not None and now - self.last_tunnel_created < self.connection_rate_limit:
             logger.info("Rejecting tunnel %s due to rate limiting" % tunnel_str)
             return False
+        if address[0] in self.last_tunnel_created_per_ip and now - self.last_tunnel_created_per_ip[address[0]] < self.connection_rate_limit_per_ip:
```
RalfJung: So the est here is simply "2 attempts from the same IP in a short time"? isn't that a bit simplistic? It might be better to make the number of attempts configurable, like "when there are 5 attempts within 20 seconds". However, that would likely also be harder to implement (I am not sure what the usual implementation strategy for this is).

broker/l2tp_broker.cfg.example
```diff
@@ -12,6 +12,8 @@ tunnel_id_base=100
 ; Reject connections if there are less than N seconds since the last connection.
 ; Can be less than a second (e.g., 0.1).
 connection_rate_limit=10
+connection_rate_limit_per_ip=1
```
RalfJung: First of all, each option here should come with a comment explaining it.
Secondly, this particular configuration makes little sense, does it? With a general rate limit of 10, an additional limit of 1 per client will never even kick in.

Also see https://github.com/wlanslovenija/tunneldigger/pull/137 where I propose to change the default settings.

broker/src/tunneldigger_broker/broker.py
RobWei: Okay. I changed this method to an adjustable one.

broker/l2tp_broker.cfg.example
RobWei: Modified in last commit.

broker/l2tp_broker.cfg.example
```diff
@@ -12,6 +12,10 @@ tunnel_id_base=100
 ; Reject connections if there are less than N seconds since the last connection.
 ; Can be less than a second (e.g., 0.1).
 connection_rate_limit=10
+; Runs broker.tunnel-connection-threshold if a IP address connect more than X times in N seconds
+connection_rate_limit_per_ip=30
+connection_rate_limit_per_ip_timescale=60
```
RalfJung, line 17:
```suggestion
connection_rate_limit_per_ip_count=30
connection_rate_limit_per_ip_times=60
```
(or similar)
Then you can write the docs as "more than COUNT times in TIME seconds", making it more clear than currently what X and N are.

broker/src/tunneldigger_broker/broker.py
```diff
@@ -37,7 +40,10 @@ def __init__(
         self.tunnel_ids = set(range(tunnel_id_base, tunnel_id_base + max_tunnels))
         self.tunnels = {}
         self.last_tunnel_created = None
+        self.last_tunnel_created_per_ip = {}
```
RalfJung, line 43: Please describe somewhere what the data stored in this dict looks like. (Gosh I wish this was written in a typed language...)

broker/src/tunneldigger_broker/broker.py
```diff
@@ -77,6 +83,26 @@ def create_tunnel(self, broker, address, uuid, remote_tunnel_id, client_features
         else:
             tunnel_str = "(%s)" % uuid
 
+        # Rate limit creation of new tunnels with the same IP address.
+        # Runs broker.tunnel-connection-threshold hook if threshold exceeded:
+        if self.connection_rate_limit_per_ip > 0 and self.connection_rate_limit_per_ip_timescale > 0:
+            if address[0] not in self.last_tunnel_created_per_ip:
+                self.last_tunnel_created_per_ip[address[0]] = deque([now])
+            if len(self.last_tunnel_created_per_ip[address[0]])-1 is self.connection_rate_limit_per_ip:
```
RalfJung, line 91: A few more comments in here would help to understand what is going on.

RalfJung:
Thanks a lot, this looks great! I like the approach, just have a few nits for the details.

@kaechele any opinion?

broker/src/tunneldigger_broker/broker.py
```diff
@@ -77,6 +83,26 @@ def create_tunnel(self, broker, address, uuid, remote_tunnel_id, client_features
         else:
             tunnel_str = "(%s)" % uuid
 
+        # Rate limit creation of new tunnels with the same IP address.
+        # Runs broker.tunnel-connection-threshold hook if threshold exceeded:
+        if self.connection_rate_limit_per_ip_count > 0 and self.connection_rate_limit_per_ip_time > 0:
+            if address[0] not in self.last_tunnel_created_per_ip:
+                self.last_tunnel_created_per_ip[address[0]] = deque([now]) # Create deque collection if IP address does not exist in dict
+            if len(self.last_tunnel_created_per_ip[address[0]])-1 is self.connection_rate_limit_per_ip_count:
```
RalfJung, line 91: Could you bind `self.last_tunnel_created_per_ip[address[0]]` to a variable? It gets used quite a bit here.

broker/src/tunneldigger_broker/broker.py
```diff
@@ -77,6 +83,26 @@ def create_tunnel(self, broker, address, uuid, remote_tunnel_id, client_features
         else:
             tunnel_str = "(%s)" % uuid
 
+        # Rate limit creation of new tunnels with the same IP address.
+        # Runs broker.tunnel-connection-threshold hook if threshold exceeded:
+        if self.connection_rate_limit_per_ip_count > 0 and self.connection_rate_limit_per_ip_time > 0:
+            if address[0] not in self.last_tunnel_created_per_ip:
+                self.last_tunnel_created_per_ip[address[0]] = deque([now]) # Create deque collection if IP address does not exist in dict
```
RalfJung, line 90: Why is this pre-filled with `now`?

broker/src/tunneldigger_broker/broker.py
RalfJung: I am also a bit confused by the `-1` here. If these are indeed all *previous* attempt, I think it should be just an equality comparison? At least, if you didn't pre-add the `now`.

broker/src/tunneldigger_broker/broker.py
```diff
@@ -77,6 +83,26 @@ def create_tunnel(self, broker, address, uuid, remote_tunnel_id, client_features
...
+            if len(self.last_tunnel_created_per_ip[address[0]])-1 is self.connection_rate_limit_per_ip_count:
+                scale_delta = now - self.last_tunnel_created_per_ip[address[0]][0] # Delta of oldest timestamp in collection and now
+                if scale_delta <= self.connection_rate_limit_per_ip_time:
+                    logger.info("Tunnel %s connection threshold exceeded (%s in %s (%s) seconds). Running hook" % (tunnel_str, self.connection_rate_limit_per_ip_count, int(scale_delta), self.connection_rate_limit_per_ip_time))
```
RalfJung, line 94: The `%s (%s)` here is not very clean.
Also, please use `"...".format()` instead of `%`-formatting.

broker/src/tunneldigger_broker/broker.py
```diff
@@ -77,6 +83,26 @@ def create_tunnel(self, broker, address, uuid, remote_tunnel_id, client_features
...
+                    broker.hook_manager.run_hook(
+                        'broker.tunnel-connection-threshold',
+                        address[0],
+                        address[1],
+                        uuid,
+                    )
+                    del self.last_tunnel_created_per_ip[address[0]]
+                    return False
+                self.last_tunnel_created_per_ip[address[0]].popleft() # Delete first (oldest) timestamp in deque collection
+            self.last_tunnel_created_per_ip[address[0]].append(now) # Append current timestamp at the end of deque collection
```
RalfJung, line 104: At this point, maybe assert that `len(self.last_tunnel_created_per_ip[address[0]])` is not too big? Your `is` above would go horribly wrong if the deque ever got too long.
I think the limit should be to never have more than `connection_rate_limit_per_ip_count` items in there, but with our `-1` I am not sure about that.

RalfJung:
The PR is still marked as WIP, so I am not sure if you are even looking for a review already?

> We will use the hook to block an IP address over iptables for 5 minutes. I have chosen the hook to allow for more flexibility in script design.

Could you add that as an example script?

broker/src/tunneldigger_broker/broker.py
```diff
@@ -77,6 +83,27 @@ def create_tunnel(self, broker, address, uuid, remote_tunnel_id, client_features
...
+        # Rate limit creation of new tunnels with the same IP address.
+        # Runs broker.tunnel-connection-threshold hook if threshold exceeded:
+        if self.connection_rate_limit_per_ip_count > 0 and self.connection_rate_limit_per_ip_time > 0:
+            if address[0] not in self.last_tunnel_created_per_ip:
+                self.last_tunnel_created_per_ip[address[0]] = deque([], self.connection_rate_limit_per_ip_count) # Create deque collection if IP address does not exist in dict
+            tunnelCollection = self.last_tunnel_created_per_ip[address[0]]
+            if len(tunnelCollection) >= self.connection_rate_limit_per_ip_count:
+                scale_delta = now - tunnelCollection[0] # Delta of oldest timestamp in collection and now
+                if scale_delta <= self.connection_rate_limit_per_ip_time:
+                    logger.info("Tunnel {0} connection threshold exceeded ({1} in {2} ({3}) seconds). Running hook".format(tunnel_str, self.connection_rate_limit_per_ip_count, int(scale_delta), self.connection_rate_limit_per_ip_time))
```
RalfJung, line 95: Please break these long lines (same above in L90).

broker/src/tunneldigger_broker/broker.py
```diff
@@ -77,6 +83,27 @@ def create_tunnel(self, broker, address, uuid, remote_tunnel_id, client_features
         else:
             tunnel_str = "(%s)" % uuid
 
+        # Rate limit creation of new tunnels with the same IP address.
+        # Runs broker.tunnel-connection-threshold hook if threshold exceeded:
+        if self.connection_rate_limit_per_ip_count > 0 and self.connection_rate_limit_per_ip_time > 0:
+            if address[0] not in self.last_tunnel_created_per_ip:
+                self.last_tunnel_created_per_ip[address[0]] = deque([], self.connection_rate_limit_per_ip_count) # Create deque collection if IP address does not exist in dict
```
RalfJung, line 90: That's cool, I did not know `deque` had a "maxlen" parameter. :D
Do you even still need the `popleft` below?

broker/src/tunneldigger_broker/broker.py
```diff
@@ -77,6 +83,32 @@ def create_tunnel(self, broker, address, uuid, remote_tunnel_id, client_features
...
+        # Runs broker.tunnel-connection-threshold hook if threshold exceeded:
+        if self.connection_rate_limit_per_ip_count > 0 and self.connection_rate_limit_per_ip_time > 0:
+            if address[0] not in self.last_tunnel_created_per_ip:
+                # Create deque collection if IP address does not exist in dict
+                self.last_tunnel_created_per_ip[address[0]] = deque([], self.connection_rate_limit_per_ip_count)
+            tunnelCollection = self.last_tunnel_created_per_ip[address[0]]
+            if len(tunnelCollection) >= self.connection_rate_limit_per_ip_count:
+                scale_delta = now - tunnelCollection[0] # Delta of oldest timestamp in collection and now
+                if scale_delta <= self.connection_rate_limit_per_ip_time:
+                    logger.info("Tunnel {0} connection threshold exceeded ({1} in {2} ({3}) seconds). Running hook".format(
```
RalfJung, line 96: I don't think "5 in 26 (30) seconds" is very comprehensible.
Also, shouldn't it be `connection_rate_limit_per_ip_count+1`? There are `connection_rate_limit_per_ip_count` previously recorded attempts, this is the next one.

broker/src/tunneldigger_broker/broker.py
```diff
@@ -77,6 +83,32 @@ def create_tunnel(self, broker, address, uuid, remote_tunnel_id, client_features
...
 
+        # Rate limit creation of new tunnels with the same IP address.
+        # Runs broker.tunnel-connection-threshold hook if threshold exceeded:
+        if self.connection_rate_limit_per_ip_count > 0 and self.connection_rate_limit_per_ip_time > 0:
+            if address[0] not in self.last_tunnel_created_per_ip:
+                # Create deque collection if IP address does not exist in dict
+                self.last_tunnel_created_per_ip[address[0]] = deque([], self.connection_rate_limit_per_ip_count)
+            tunnelCollection = self.last_tunnel_created_per_ip[address[0]]
+            if len(tunnelCollection) >= self.connection_rate_limit_per_ip_count:
+                scale_delta = now - tunnelCollection[0] # Delta of oldest timestamp in collection and now
```
RalfJung, line 94: What is being "scaled" here? I do not understand this variable name.

RalfJung:
@RobWei did you see my comment at https://github.com/wlanslovenija/tunneldigger/pull/138#issuecomment-611932677?

Also, this PR has merge conflicts with master.

And finally, should we be worried about accumulating too much data in that new dict you are adding? Is it worth having periodic cleanup, deleting all entries where the last connection attempt is more than TIME ago so it cannot be relevant any more?

RalfJung:
@RobWei seems to be busy with other stuff and we are also seeing connection storms from single IPs in our network, so I'll likely push this PR to completion myself soon.

RobWei:
@RalfJung  
I am currently in my project work and still have 5 exam parts to pass.

I would be happy if someone can complete it, otherwise I would have to look at the Pull Request again in about one and a half months.

RalfJung:
@RobWei understood, thanks for letting me know! I'll add this to my list.

Good luck with your exams. :)

kaechele:
I am unsure whether we should use the IP and not the UUID as a key here.
Multiple clients could be behind the same IP address (think DS-Lite with CGNAT) and this change would punish regular clients if one client behind the same IP misbehaves.
I guess implementing this makes sense but the UUID as a key should also be implemented and the aforementioned caveat should be documented accordingly.

RalfJung:
We can use the IP-UUID tuple, and see if that helps. If it does not we can always refine later.

kaechele:
Actually I just re-read what this PR does and I find the description slightly misleading.
This merely provides a hook that's being called when a client exceeds the rate limit. No rate-limiting is actually being done here.
Effectively this sounds more like a hack as it does not solve the actual problem, namely clients reconnecting too fast. Implementing some kind of back-off on the client side would be needed to fix the actual problem.
I am not in favour of providing a band-aid when the actual problem could be solved.
I do agree that it is necessary though, we are seeing the same issues here.

RalfJung:
I agree there is a bug in the client. However, I have no idea what is happening in the client to cause this. The client shouldn't connect to many brokers in a tight loop like these here do. Right now I don't even know where to start fixing this on the client.

RalfJung:
@RobWei can you share the hook script that you are using to block clients, the one that sets up an iptables rule for 5 minutes?

RalfJung:
I rebased this and updated it to take my comments into account. It is running fine on at least one of our servers (but without a hook configured).

But it would be nice to have an example hook as well.

RalfJung:
Curious, CI is failing (and likely not spurious). Probably it triggers the rate limits.^^

RalfJung:
This keeps getting weirder...
```
ERROR: Invalid signature for /tmp/tmp.WR47aol82I/index.asc
```

RalfJung:
Thanks a lot @RobWei for the initial implementation. :)

FYI, I renamed the hook in the configuration, so if you switch to master you may have to update your config file.

Issue #144, opened 2020-07-24
Change rate limiting to be per-UUID


RalfJung:
@RobWei recently contributed per-IP rate limiting. I was excited to try and see if this helps mitigate https://github.com/wlanslovenija/tunneldigger/issues/143, but unfortunately that is not the case: the bad node reconnects with a rate that is too low for reasonable rate limiting, in particular when considering that people could run multiple nodes under the same IP.

Hence I propose to change that rate-limiting to be per-UUID instead of per-IP. The motivating log in https://github.com/wlanslovenija/tunneldigger/pull/138 shows the same UUID over and over, so this approach should still help @RobWei, but at the same time it means we can set the rate limit much lower because we do not have to worry about the same IP being used by multiple clients.

RalfJung:
Closing as https://github.com/wlanslovenija/tunneldigger/pull/148 fixes the underlying load-causing problem.

Issue #143, opened 2020-07-24
High CPU load due to a single misbehaving client


RalfJung:
We occasionally see a client misbehave and establish multiple connections at the same time to all our servers. For some reason, even when there are just around 20 connections per 10 minutes, this causes 100% CPU load by tunneldigger. Python is not the most efficient language, but this seems eccessive -- I'd like to better understand where in the broker all that CPU time is spent. Unfortunately, so far I found no good way to do such an analysis for python (what I am looking for is something like callgrind).

RalfJung:
I did a `cProfile` run of this (on the live system under the problematic load situation), so now I can start analyzing that profiler data. So far I am not sure what to conclude.

Here's the functions with the highest "cumtime":
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        2    0.000    0.000  409.537  204.769 main.py:1(<module>)
     51/1    0.000    0.000  204.769  204.769 {built-in method builtins.exec}
     71/1    0.000    0.000  204.769  204.769 <frozen importlib._bootstrap>:978(_find_and_load)
     71/1    0.000    0.000  204.769  204.769 <frozen importlib._bootstrap>:948(_find_and_load_unlocked)
    104/2    0.000    0.000  204.769  102.384 <frozen importlib._bootstrap>:211(_call_with_frames_removed)
     68/2    0.000    0.000  204.768  102.384 <frozen importlib._bootstrap>:663(_load_unlocked)
     47/2    0.000    0.000  204.768  102.384 <frozen importlib._bootstrap_external>:722(exec_module)
        1   79.314   79.314  203.672  203.672 eventloop.py:44(start)
 60789078  112.203    0.000  112.203    0.000 {method 'poll' of 'select.epoll' objects}
 84148432   11.585    0.000   11.585    0.000 {method 'get' of 'dict' objects}
       22    0.002    0.000    1.448    0.066 tunnel.py:239(close)
       91    0.001    0.000    1.392    0.015 netlink.py:127(send)
       91    0.000    0.000    1.392    0.015 netlink.py:152(send)
       91    1.391    0.015    1.391    0.015 {method 'send' of '_socket.socket' objects}
       22    0.001    0.000    1.388    0.063 l2tp.py:181(session_delete)
        1    0.000    0.000    1.048    1.048 broker.py:194(close)
      481    0.002    0.000    0.421    0.001 network.py:88(read)
[...]

PR #146, opened 2020-08-08
switch per per-IP to per-UUID rate limiting

RalfJung:
I rolled out per-IP rate limiting in our network, but it turns out that even rates as low as 20 connections per 10 minutes are problematic. I am not sure what exactly these misbehaving clients are doing; probably the actual issue is not the load caused by these connection attempts but by other things they do.

Doing the limiting per-UUID lets me reduce the threshhold to something like "max 5 connections per 2 minutes". I would not want to do that per-IP, because people might run 5 nodes behind the same IP and then that limit is reached too easily.

Cc @RobWei does per-UUID rate limiting work for you, or do you require a per-IP limit?

Fixes https://github.com/wlanslovenija/tunneldigger/issues/144

### DIFF
--- a/broker/l2tp_broker.cfg.example
+++ b/broker/l2tp_broker.cfg.example
@@ -14,11 +14,11 @@ tunnel_id_base=100
 ; Reject connections if there are less than N seconds since the last connection.
 ; Can be less than a second (e.g., 0.1).
 connection_rate_limit=0.2
-; Reject connection if an IP address connects more than COUNT times in TIME seconds. Also runs
+; Reject connection if a UUID connects more than COUNT times in TIME seconds. Also runs
 ; "broker.connection-rate-limit" hook (e.g. to block client via iptables).
 ; Disabled when at least one value is 0 (the default).
-;connection_rate_limit_per_ip_count=20
-;connection_rate_limit_per_ip_time=60
+;connection_rate_limit_per_uuid_count=20
+;connection_rate_limit_per_uuid_time=60
 ; Set PMTU to a fixed value.  Use 0 for automatic PMTU discovery.  A non-0 value also disables
 ; PMTU discovery on the client side, by having the server not respond to client-side PMTU
 ; discovery probes.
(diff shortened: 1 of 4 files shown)


mitar:
What about some way to collects some stats per UUID? To get to determine what are those clients doing? I worry that if they are malicious clients doing something on purpose, they would just rotate UUIDs.

RalfJung:
Sure they would. If they are malicious they could do tons of things, I don't think tunneldigger is very well prepared to deal with that. We currently have DoS problems through likely non-malicious misbehaving clients (https://github.com/wlanslovenija/tunneldigger/issues/143) which this PR helps handle; that says to me that we are quit far away from being resistant against malicious clients.

RalfJung:
With https://github.com/wlanslovenija/tunneldigger/pull/148 we do not have the load problems any more. Absent that, I agree that per-IP rate limiting makes most sense, so let's close this.

mitar:
> We currently have DoS problems through likely non-malicious misbehaving clients

I hope you are right that it is just non-malicious.

> that says to me that we are quit far away from being resistant against malicious clients.

Yea, that was not even original intent to make it so. For that we always assumed that security is done on another level (IPSec).

> we do not have the load problems any more

I am glad to hear that.

RalfJung:
> For that we always assumed that security is done on another level (IPSec).

Well, that's not really possible for Freifunk because we run open infrastructure...
(Also I don't think the docs ever said that, so when switching to tunneldigger there was no way we could have been aware of this. Now it's a bit too late with several big Freifunk networks relying on tunneldigger already.^^)

mitar:
It didn't say because it was not the goal of the project. Same as you wrote, it is for open infrastructure. Not to mention that IPSec would add some overhead. So yea, in theory you should make that using IPSec, but given our networks design, we probably do not want that, but that opens a window to malicious attacks. I do not think there is much one can do about that, it is a conscious design choice.

PR #147, opened 2020-08-08
print warning when we run out of tunnel IDs

RalfJung:


broker/src/tunneldigger_broker/broker.py
```diff
@@ -123,6 +123,7 @@ def create_tunnel(self, broker, address, uuid, remote_tunnel_id, client_features
         try:
             tunnel_id = self.tunnel_ids.pop()
         except KeyError:
+            logger.warning("No more tunnel IDs available -- {} active tunnels".format(len(self.tunnels)))
```
mitar, line 126: Just a nitpick. Python logging you should use `%` based interpolation. Moreover, you should not interpolate directly, but pass things as a secondary argument and interpolation would happen only if the logging level forces the message to be printed out. Now you are interpolating always. Moreover, you can then send whole logging records to another server for logging, and not just interpolated strings.

So this should be something like:

```
logger.warning("No more tunnel IDs available -- %(active_tunnels)s active tunnels", {'active_tunnels': len(self.tunnels)})
```

The one above:

```
logger.info("Rejecting tunnel %(tunnel_str)s due to per-IP rate limiting: %(attempt_count)s attempts in %(interval)s seconds", {
    'tunnel_str': tunnel_str,
    'attempt_count': len(tunnelCollection),
    'interval': int(delta),
})
```

RalfJung: Hm I thought I have read once that `%` interpolation is basically deprecated. But maybe the logger API is older than that, or so. I'll adjust this.

RalfJung: `%` strings are quite annoying though as I have to repeat the type of the expression being printed... there's a reason modern languages do not use them any more. :/

RalfJung: Also note that the rest of the codebase uses explicit `%` interpolation, so it is not lazy either.

mitar: > Hm I thought I have read once that % interpolation is basically deprecated.

It is.

> But maybe the logger API is older than that, or so.

Yea. That is the reason. They have a section how to [address this](https://docs.python.org/3/howto/logging-cookbook.html#formatting-styles), but it is really ugly in my view. And so I think it is just easiest to use this old format. But given that Tunneldigger is stand-alone program maybe some of ways mentioned there would work. The issue here is that it is hard to do it for libraries, but if you control whole codebase, then it can work, if you want to configure alternative formatting style.

mitar: > Also note that the rest of the codebase uses explicit % interpolation, so it is not lazy either.

Yes, I just noticed it here, so this is why mentioned it. It is not a major thing anyway. I just wanted to share because in one other project I am using Python logging a lot and using this feature to direct logs to a central server, not as strings, but as log records (so log message + dict with parameters separate), and then have rules there to detect error conditions. It is easier this way because you do not have to parse things out of the string again.

mitar: > strings are quite annoying though as I have to repeat the type of the expression being printed...

In Python not really. I just always use `%(...)s` everywhere and it just generally does the right thing, unless you want less digits or something in a particular number.

mitar: Also, it is better to use keyword-based interpolation than position one. Less error prone and more metadata what each value is.

RalfJung: > I just wanted to share because in one other project I am using Python logging a lot and using this feature to direct logs to a central server, not as strings, but as log records (so log message + dict with parameters separate), and then have rules there to detect error conditions. It is easier this way because you do not have to parse things out of the string again.

Oh that's really cool. :D

> In Python not really. I just always use %(...)s everywhere and it just generally does the right thing

Ah so `%s` works for integers and floats? That's good to know.

> Also, it is better to use keyword-based interpolation than position one. Less error prone and more metadata what each value is.

Tunneldigger uses positional stye everywhere so I don't think this PR is the right time to change coding style.

mitar: > Oh that's really cool. :D

E.g., https://docs.python.org/3/howto/logging-cookbook.html#sending-and-receiving-logging-events-across-a-network

> Ah so %s works for integers and floats? That's good to know.

Yes. It is like running `str(...)` on the value. So in most cases this is what you want anyway. Only on float you might get too many digits which makes it hard to read.

> Tunneldigger uses positional stye everywhere so I don't think this PR is the right time to change coding style.

Agreed.

RalfJung: > Agreed.

So this PR is read to be merged then?

mitar: Yes, of course. :-)

PR #148, opened 2020-08-09
close pollable when its FD errors

RalfJung:
When an FD errors, `epoll` returns immediately reporting the error. This leads to a tight loop (instead of epoll waiting until there is a new event), causing very high load (https://github.com/wlanslovenija/tunneldigger/issues/143). So we should stop polling FDs once they error. This PR achieves that by simply closing the associated pollable.

I checked that on our server with a misbehaving client whose connections keep dropping this reduces the load from 100% to basically nothing.

Fixes https://github.com/wlanslovenija/tunneldigger/issues/143.

### DIFF
--- a/broker/src/tunneldigger_broker/eventloop.py
+++ b/broker/src/tunneldigger_broker/eventloop.py
@@ -57,7 +57,7 @@ class EventLoop(object):
 
                     if event & select.EPOLLIN:
                         pollable.read(file_object)
-                    elif event & select.EPOLLERR:
+                    elif event & select.EPOLLERR or event & select.EPOLLHUP:
                         pollable.close()
             except IOError:
                 # IOError get produced by signal even. in version 3.5 this is fixed an the poll retries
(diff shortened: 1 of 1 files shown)


broker/src/tunneldigger_broker/eventloop.py
```diff
@@ -57,6 +57,8 @@ def start(self):
 
                     if event & select.EPOLLIN:
                         pollable.read(file_object)
+                    elif event & select.EPOLLERR:
```
mitar, line 60: You might want to handle `EPOLLHUP` as well here, if you are adding this: https://stackoverflow.com/questions/24119072/how-to-deal-with-epollerr-and-epollhup

`EPOLLHUP` is when the connection is closed nicely. It [looks like that those two events are always listened for](https://man7.org/linux/man-pages/man2/epoll_ctl.2.html).

RalfJung: Good catch! I made it also close the pollable on HUP.

kaechele:
LGTM. We were seeing this issue as well.

RalfJung:
All right, merging then.

PR #150, opened 2020-08-10
handle hook process being closed twice

RalfJung:
After merging https://github.com/wlanslovenija/tunneldigger/pull/148, I started seeing exceptions terminating the broker on one (the slowest) of our servers. Seems like reacting to EPOLLHUP means we now close some things twice, which the `HookProcess` class did not handle well. This fixes the exceptions, but I am not sure if it is a proper fix.

### DIFF
--- a/broker/src/tunneldigger_broker/hooks.py
+++ b/broker/src/tunneldigger_broker/hooks.py
@@ -55,6 +55,10 @@ class HookProcess(object):
         Closes the hook process.
         """
 
+        if not hasattr(self, "buffer"):
+            # We have already been closed.
+            return
+
         for line in self.buffer.getvalue().decode('utf-8').split('\n'):
             if not line:
                 continue


broker/src/tunneldigger_broker/hooks.py
```diff
@@ -55,6 +55,10 @@ def close(self):
         Closes the hook process.
         """
 
+        if not hasattr(self, "buffer"):
```
mitar, line 58: I think it would be better to have an explicit `is_closed` attribute on the object so that we can immediately set it the first thing in `close`. Otherwise there might be some race conditions.

RalfJung: That's still not atomic, so it will not solve race conditions. But also, this is all sequential, right?

mitar: Probably true.

RalfJung:
We've been running this in production for a few days now and it all seems fine. Any objections to merging this?

mitar:
Not from me.

PR #151, opened 2020-09-27
actually seed randomness if we use it

RalfJung:
Original patch by Peter J. Philipp, I adjusted it a bit.

### DIFF
--- a/client/l2tp_client.c
+++ b/client/l2tp_client.c
@@ -1354,7 +1354,13 @@ int main(int argc, char **argv)
       }
       case 'a': select_broker = broker_selector_usage; break;
       case 'g': select_broker = broker_selector_first_available; break;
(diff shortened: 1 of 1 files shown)


mitar (approved):

mitar:
BTW, on clients clock might not really be too random. Maybe we should seed randomness from some system device? At least there you can store entropy to preserve it across restarts.

RalfJung:
I don't think it is very important that the broker selection uses cryptographic-strength randomness.^^

mitar:
Oh, I am not talking about cryptographic-strength randomness. Only what happens if in some area electricity goes out, then comes back in, and then all nodes "randomly" pick the same server, because they have all same time. :-)

But sure. :-)

RalfJung:
Fair.^^ That's why I used nanoseconds, I hoped those are sufficiently noisy...

PR #153, opened 2020-10-01
Made interface to an optional parameter

RobWei:
Since we are now running a tunneligger broker instance on a routed IP address, I noticed that the tunneligger does not work with a loopback interface.

To the basic structure:
Our IPv4 subnet is announced via BGP on several interfaces. An IP from the subnet is configured on the loopback interface.

A short analysis of the pollable object showed that the UDP socket is bound to an interface.  I could not find out why this was done.

This change allows to define interfaces optionally.

### DIFF
--- a/broker/src/tunneldigger_broker/main.py
+++ b/broker/src/tunneldigger_broker/main.py
@@ -91,11 +91,18 @@ brokers = []
 broker_host = config.get('broker', 'address')
 for port in config.get('broker', 'port').split(','):
     try:
-        broker_instance = broker.Broker(
-            (broker_host, int(port)),
-            config.get('broker', 'interface'),
-            tunnel_manager,
-        )
+        if config.has_option('broker', 'interface'):
+            broker_instance = broker.Broker(
+                (broker_host, int(port)),
+                config.get('broker', 'interface'),
+                tunnel_manager,
+            )
+        else:
+            broker_instance = broker.Broker(
+                (broker_host, int(port)),
+                None,
+                tunnel_manager,
+            )
         logger.info("Listening on %s:%d." % broker_instance.address)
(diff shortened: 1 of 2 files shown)


mitar:
Have you tested this? CI is failing.

RalfJung:
The CI failure is likely our spurious LXC issue -- https://github.com/lxc/lxc/issues/3467.

broker/src/tunneldigger_broker/main.py
```diff
@@ -91,11 +91,18 @@
 broker_host = config.get('broker', 'address')
 for port in config.get('broker', 'port').split(','):
     try:
-        broker_instance = broker.Broker(
-            (broker_host, int(port)),
-            config.get('broker', 'interface'),
-            tunnel_manager,
-        )
+        if config.has_option('broker', 'interface'):
```
RalfJung, line 94: Changing whether an option is optional should be reflected in the docs and/or example config file as well.

RalfJung:
> Have you tested this? Can you add a test case to the CI for this to be tested to work without interface and IP bound to localhost?

@mitar  Our CI is quite hard to navigate, what did you imagine here?
Currently the broker config for CI is [set up here](https://github.com/wlanslovenija/tunneldigger/blob/c187581a702c4cf060b966169c651d8a7a568046/tests/prepare_server.sh#L57), I guess we could remove the `interface` option for some of the CI runs?

mitar:
My main worry is that we are changing something here and we are not sure why it was required in the first place. Is this related to NAT stuff which was in there in the past? In any case, making this optional is a new feature which we should make sure that it a) really works b) keeps working. I am currently not convinced into any of those. Is interface really optional? Do things really keep working without it?

PR #155, opened 2020-10-18
Update server.rst

peterschristoph:
add more details for setup under debian 10 buster (with kernel < 5)

### DIFF
--- a/docs/server.rst
+++ b/docs/server.rst
@@ -98,9 +98,19 @@ your listing of the modules might look like this:
 System Packages
 ---------------
 
-Also the following Debian packages are required:
+Also the following Debian <= 9 packages are required:
 
 * ``iproute``
+* ``bridge-utils``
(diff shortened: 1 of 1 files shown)


docs/server.rst
```diff
@@ -122,25 +132,30 @@ not affect the versions that are installed for Tunneldigger.
...
+
 If we assume that you are installing Tunneldigger under ``/srv/tunneldigger``
 (the scripts provided with Tunneldigger assume that as well), you can do::
 
     cd /srv/tunneldigger
     virtualenv -p /usr/bin/python3 env_tunneldigger
 
 .. note::
-    Tunneldigger only supports Python 3.
+    Tunneldigger in the newest version only supports Python 3.
```
peterschristoph, line 158: I'm not sure since when. But there are definitely tunneldigger versions e.g. v0.3.0 that run with python 2.7

Issue #145, opened 2020-07-30
No releases since 2017?


PolynomialDivision:
Why is no release released since 3 years? :)

RalfJung:
I just use the latest git version. Development is slow enough that a release just doesn't feel worth it.

mitar:
He he. But there are some major changes though, like Python 3 support, session ID and stuff.

I think maybe a new version release is in order, it is easier to pack then upstream, because packages generally depend on released versions.

RalfJung:
I want to tweak our rate limiting some more. Once that's in, I'm fine with a release, but I do not think I will have the time to do the work for that.

PR #156, opened 2020-10-18
Update server.rst

peterschristoph:
add debian 10 requirements, for easier setup.

on debian 10 the system packet "libnetfilter-conntrack-dev" is definitely missing in the docs otherwise the installation will abort at the legacy branch.

### DIFF
--- a/docs/server.rst
+++ b/docs/server.rst
@@ -115,6 +115,13 @@ Also the following Debian packages are required:
 * ``python-dev``
 * ``libevent-dev``
 
+Debian 10 required following packages:
+* ``iproute2``
+* ``bridge-utils``
+* ``python-dev``
+* ``libevent-dev``
+* ``libnetfilter-conntrack-dev``
+
 If you would like to use the already supplied hook scripts to setup the network
 interfaces, you also need the following packages:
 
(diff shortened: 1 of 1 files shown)


docs/server.rst
```diff
@@ -147,7 +158,7 @@ If we assume that you are installing Tunneldigger under ``/srv/tunneldigger``
     virtualenv -p /usr/bin/python3 env_tunneldigger
 
 .. note::
-    Tunneldigger only supports Python 3.
+    Latest Tunneldigger only supports Python 3.
```
peterschristoph, line 161: It's confusing, because the "stable" version shown as v0.3.0 is working under python 2.7.

Maybe someone can delete the "stable" read the docs version or update this? how can something be stable if it runs under python 2.7?

mitar: I think this is related to https://github.com/wlanslovenija/tunneldigger/issues/145?

RalfJung: I don't think docs should retrace the history of all things that used to be supported in the past... *everything* the docs say only applies to latest tunneldigger. It would be more confusing to emphasize this here as that might be read to imply that some of the rest of the docs also apply to older tunneldigger, which they do not.

docs/server.rst
```diff
@@ -115,6 +115,13 @@ Also the following Debian packages are required:
 * ``python-dev``
 * ``libevent-dev``
 
+Debian 10 required following packages:
```
RalfJung, line 118: If we give a distro here, we should also give a distro above -- or replace the above list.

docs/server.rst
```diff
@@ -115,6 +115,13 @@ Also the following Debian packages are required:
...
+* ``iproute2``
+* ``bridge-utils``
+* ``python-dev``
+* ``libevent-dev``
+* ``libnetfilter-conntrack-dev``
```
RalfJung, line 123: This packet is not required any more as far as I know -- we do not have it installed on our servers and tunneldigger works fine there.

PR #157, opened 2020-10-19
show more details on connection close

RalfJung:


broker/src/tunneldigger_broker/tunnel.py
```diff
@@ -243,7 +243,7 @@ def close(self, reason=protocol.ERROR_REASON_UNDEFINED):
         :param reason: Reason code for the tunnel being closed
         """
 
-        logger.info("Closing tunnel %d (%s) after %d seconds", self.tunnel_id, self.uuid, time.time() - self.created_time)
+        logger.info("Closing tunnel {} ({}) after {} seconds (reason=0x{:x})".format(self.tunnel_id, self.uuid, int(time.time() - self.created_time), reason))
```
mitar, line 246: Nit: Don't pre-format in logging. :-)

RalfJung: I know you prefer the `%` style but hey formatting is so much easier with the modern syntax...

RalfJung: I know you prefer the % style but hex formatting is so much easier with the modern syntax...

mitar: True. :-)

You could also do:

```
logger.info("Closing tunnel %d (%s) after %d seconds (reason=%s)", self.tunnel_id, self.uuid, time.time() - self.created_time, "0x{:x}".format(reason))
```

mitar:
BTW, "This branch is out-of-date with the base branch".

mitar (approved):

broker/src/tunneldigger_broker/tunnel.py
```diff
@@ -330,6 +330,7 @@ def message(self, address, msg_type, msg_data, raw_length):
         if msg_type == protocol.CONTROL_TYPE_ERROR:
             # Error notification from the remote side.
             # TODO: Parse error code.
+            logger.warning("Tunnel %d (%s) got error from remote peer", self.tunnel_id, self.uuid)
```
mitar, line 333: What about parsing the error code and logging it?

RalfJung: I mean, that would be nice, but for now having the code printed is better than nothing I think.

Issue #158, opened 2020-11-20
Review CI Tooling


kaechele:
In the light of Travis CI restricting use of their free tier (see https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) we may want to think about re-tooling our CI pipeline.

RalfJung:
We are using travis-ci.org though (the OSS version), not travis-ci.com (the commercial version that provides a free tier to OSS sometimes). It is unclear to me what is changing there, since the blog post above emphasized to be about travis-ci.com.

kaechele:
The .org product is being shut down December 31 and users migrated to the .com one: https://mailchi.mp/3d439eeb1098/travis-ciorg-is-moving-to-travis-cicom

RalfJung:
Ah, fun. You'd think they'd email people actively using travis-ci.org that the service is being discontinued, but it seems they consider that unnecessary...

Time to move to GitHub Actions, it seems.

mitar:
Hm, I think we could also just ask for OSS minutes. They say:

> We will be offering an allotment of OSS minutes that will be reviewed and allocated on a case by case basis. Should you want to apply for these credits please open a request with Travis CI support stating that you’d like to be considered for the OSS allotment.

I think it is pretty clear we are an open source project here which does not plan to misuse them.

> Time to move to GitHub Actions, it seems.

From my understanding, I would doubt this it is really possible to replicate our current setup with multiple docker containers trying to test networking between them.

I have had good success with GitLab CI, [which you can also use just as CI for GitHub only](https://about.gitlab.com/solutions/github/). You can run Docker in Docker there and use that to test things we might need here. So I am suggesting that if porting to GitHub Actions fails, maybe GitLab CI is to try.

RalfJung:
> Hm, I think we could also just ask for OSS minutes. They say:

That would also be an option, yeah. And then we wouldn't have to reconfigure everything.^^

Does anyone want to take care of this?

mitar:
OK, I looked into instructions a bit, to send them an e-mail and I realized we have first to do our homework: we should move CI to travis.com, get free plan, and see how long the current minutes last us. Then we should see how much we are using per month, and then we can write to them to ask for that amount to be given us monthly.

Can somebody do this transfer and see how much minutes we need at all?

RalfJung:
[...]

PR #159, opened 2020-11-30
README: point to new CI

RalfJung:
Also let's see if using travis-ci.com works.

Fixes https://github.com/wlanslovenija/tunneldigger/issues/158.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -3,8 +3,8 @@ Tunneldigger
 
 L2TPv3 VPN tunneling solution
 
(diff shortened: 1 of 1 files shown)


RalfJung:
@mitar Looks like we'll need to update branch protection as well, but so far it does not offer the new status reported by travis-ci.com.

RalfJung:
Ah, I just had to wait a bit. Now it seems to work.

PR #162, opened 2021-03-18
docs: add a preliminary protocol description

lynxis:
Describe the protocol using packetdiag.


docs/conf.py
```diff
@@ -25,7 +25,7 @@
 
 # Add any Sphinx extension module names here, as strings. They can be extensions
 # coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
-extensions = ['sphinx.ext.intersphinx']
+extensions = ['sphinx.ext.intersphinx', 'sphinxcontrib.packetdiag']
```
RalfJung, line 28: What does this change do?

docs/protocol.rst
```diff
@@ -0,0 +1,60 @@
...
+The L2TPv3 is specified in `RFC3931 <https://tools.ietf.org/html/rfc3931>`_
+
+Control packets
+***************
+
+Tunneldigger is not following RFC3931 for control messages.
+Tunneldigger is only supporting the T bit in the first bit to
+mark a packet as control data. All other fields of RFC3931
+are ignore and have a different meaning..
+All fields are encoding as network byte order.
```
RalfJung, line 18:
```suggestion
All fields are encoded with network byte order.
```

docs/protocol.rst
```diff
@@ -0,0 +1,60 @@
...
+        24-31:  Version
+        32-39:  Type
+        40-47:  Length
+        48-63:  Value
+    }
+
+
+* The T bit must be 1
+* Version must be 1
+* Type of the PDU
```
RalfJung, line 38: What's a "PDU"?

docs/protocol.rst
```diff
@@ -0,0 +1,60 @@
...
+        colwidth = 16
+        node_height = 48
+
+        0:      T
+        1-7:    0
+        8-23:   Magic 0x73A7
+        24-31:  Version
+        32-39:  Type
+        40-47:  Length
+        48-63:  Value
```
RalfJung, line 32: The length of value depends on the type; why do you write `48-63` here?

docs/protocol.rst
RalfJung: Tunneldigger calls these CONTROL_TYPE or msg_type; that might be more suited terminology.

docs/capture_analyse_packets.rst
```diff
@@ -0,0 +1,52 @@
...
+There is a `custom dissector <https://github.com/wlanslovenija/tunneldigger/blob/master/docs/wireshark-tunneldigger.lua>`_ for tunneldigger written in lua. The dissector is registered as **TD**.
+
+To use the wireshark dissector call wireshark with:
+
+.. code:: sh
+
+   cd tunneldigger/docs/
+   wireshark -Xlua_script:wireshark-tunneldigger.lua
+
+Wireshark might decode the user data as a different protocol (e.g. Cisco HDLC). This can be changed by:
```
RalfJung, line 46: It might be worth also explaining how to change the port; tunneldiger is often run on a port different than 8942.

docs/protocol.rst
lynxis: Protocol data unit https://en.wikipedia.org/wiki/Protocol_data_unit

RalfJung: I see. I don't think we use this term anywhere in tunneldigger though?

docs/conf.py
lynxis: Allows to use packet diagrams. http://blockdiag.com/en/nwdiag/packetdiag-examples.html

RalfJung: Oh, that looks neat. :)

docs/protocol.rst
lynxis: Ack

docs/protocol.rst
lynxis: Because it's a packet diagraph. I would like to have a value in there. It's only an example.

RalfJung: The problem is that it is not at all clear this is just an example. And that makes it very confusing IMO.

PR #165, opened 2022-04-10
update sample scripts to use iproute

kaechele:
Old `ioctl` based tools like `brctl` are deprecated and have been
removed from the default package set of some distributions.

Also drop usage of ebtables in favour of native bridge port isolation
available in kernels 4.18 and newer.

### DIFF
--- a/broker/scripts/bridge_functions.sh
+++ b/broker/scripts/bridge_functions.sh
@@ -7,7 +7,7 @@ ensure_policy()
 ensure_bridge()
 {
   local brname="$1"
-  brctl addbr $brname 2>/dev/null
+  ip link add $brname type bridge 2>/dev/null
   
   if [[ "$?" == "0" ]]; then
     # Bridge did not exist before, we have to initialize it
@@ -16,8 +16,6 @@ ensure_bridge()
     ip addr add 10.254.0.2/16 dev $brname
     # TODO Policy routing should probably not be hardcoded here?
     ensure_policy from all iif $brname lookup mesh prio 1000
-    # Disable forwarding between bridge ports
-    ebtables -A FORWARD --logical-in $brname -j DROP
   fi
 }
 
--- a/broker/scripts/session.mtu-changed.sh
+++ b/broker/scripts/session.mtu-changed.sh
@@ -7,12 +7,8 @@ NEW_MTU="$5"
 . scripts/bridge_functions.sh
 
 # Remove interface from old bridge
-brctl delif digger${OLD_MTU} $INTERFACE
+ip link set dev $INTERFACE nomaster
 
-# Change interface MTU
-ip link set dev $INTERFACE mtu $NEW_MTU
-
-# Add interface to new bridge
+# Change interface MTU and add to new bridge
 ensure_bridge digger${NEW_MTU}
-brctl addif digger${NEW_MTU} $INTERFACE
-
+ip link set dev $INTERFACE master digger${NEW_MTU} mtu $NEW_MTU up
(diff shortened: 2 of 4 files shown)


broker/scripts/session.up.sh
```diff
@@ -11,10 +11,9 @@ LOCAL_BROKER_PORT="$9"
...
-# Set the interface to UP state
-ip link set dev $INTERFACE up mtu $MTU
-
 # Add the interface to our bridge
 ensure_bridge digger${MTU}
-brctl addif digger${MTU} $INTERFACE
+ip link set dev $INTERFACE master digger${MTU} mtu $MTU up
 
+# Turn on bridge port isolation
+bridge link set dev $INTERFACE isolated on
```
RalfJung, line 19: This looks nicer than ebtables indeed, but OTOH there is a race condition here now, is there? Between the time this is added to the bridge, and when this `bridge link set` is executed, the host *can* communicate with all other hosts?

kaechele: Thanks for pointing this out. I think you might be right. Maybe we can `up` the link only after setting `isolated on`. I'll test that.

pmelange: Would a similar "isolated on" be needed in the mtu_changed script?

pmelange: I just checked, and yes, it would be necessary to set to isolated again when changing bridges with the mtu script.  The port status can be seen in /sys/class/net/$INTERFACE/brport/isolated

Also, it's not necessary to use the bridge command.  You can also simply ```echo 1 > /sys/class/net/$INTERFACE/brport/isolated```

kaechele: Pushed a fix to my branch. Thanks for testing.

broker/scripts/bridge_functions.sh
```diff
@@ -16,8 +16,6 @@ ensure_bridge()
     ip addr add 10.254.0.2/16 dev $brname
     # TODO Policy routing should probably not be hardcoded here?
     ensure_policy from all iif $brname lookup mesh prio 1000
-    # Disable forwarding between bridge ports
-    ebtables -A FORWARD --logical-in $brname -j DROP
```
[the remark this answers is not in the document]
pmelange: You have to do the isolation per port, and isolating the bridge interface itself would mean that none of the attached bridge porst would be able to communicate with host.

broker/scripts/session.mtu-changed.sh
```diff
@@ -7,12 +7,14 @@ NEW_MTU="$5"
...
+# Change interface MTU and add to new bridge
 ensure_bridge digger${NEW_MTU}
-brctl addif digger${NEW_MTU} $INTERFACE
+ip link set dev $INTERFACE master digger${NEW_MTU} mtu $NEW_MTU
+
+# Turn on bridge port isolation
+bridge link set dev $INTERFACE isolated on
 
+# Bring the tunnel interface up only after port isolation is enabled
+ip link set dev $INTERFACE up
```
pmelange, line 20: I believe that just removing the interface from the old bridge does not bring the interface down.  So, I suggest just after removing the interface from the old bridge, do an ```ip link down```

broker/scripts/bridge_functions.sh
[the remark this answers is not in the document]
kaechele: @pmelange is correct. `isolation on` is a function of the bridge port, not the bridge itself. As such it would functionally belong into the session handler scripts.

broker/scripts/session.mtu-changed.sh
kaechele: I will need to test this. Conversely, if what you say is true we could, things should continue to work even if we don't bring _up_ the interface as we add it to the bridge for the new MTU value.

PR #172, opened 2023-08-15
when the broker shuts down, give a reason for closing the tunnels

RalfJung:


RalfJung:
Ah dang, we don't have working CI any more... it needs to be ported to GHA...

PR #184, opened 2023-08-16
move socket read error handling logic around a bit

RalfJung:
We always `read` when epoll says there is an error waiting. Then `read` will dispatch to `error` to get handling suitable for the specific socket. For `Tunnel` we avoid doing anything (logging or counting the error) for EMSGSIZE.

I realized we are entirely ignoring write errors, so I made them feed the same mechanism.

@mitar is that along the lines of what you had in mind?


mitar:
Yea. This looks good.

PR #192, opened 2025-11-09
client: set cmake_minimum_required to 3.10

PolynomialDivision:
New cmake versions require at least 3.5 as 'cmake_minimum_required' in CMakeLists.txt. In future 3.10 will be required.


PolynomialDivision:
@RalfJung Can you make a new release so we can update openwrt version and dont need to carry a patch?

RalfJung:
Uh, can't you just take a git commit? I have zero recollection of what we did for the last release, and I am not sure that's the best use of the little time I can invest in this project...

